PR #186, opened 2020-10-29
cli: Improve file operation error messages

bercarug:
Improves the error messsages displayed when a file
operation fails, by outputting the file name as well
as the attempted operation.

### DIFF
--- a/src/enclave_proc/cpu_info.rs
+++ b/src/enclave_proc/cpu_info.rs
@@ -11,6 +11,9 @@ use crate::common::commands_parser::RunEnclavesArgs;
 use crate::common::{NitroCliErrorEnum, NitroCliFailure, NitroCliResult};
 use crate::new_nitro_cli_failure;
 
+/// Path corresponding to the NE CPU pool.
+const POOL_FILENAME: &str = "/sys/module/nitro_enclaves/parameters/ne_cpus";
+
 /// The CPU configuration requested by the user.
 #[derive(Clone, PartialEq)]
 pub enum EnclaveCpuConfig {
@@ -157,13 +160,13 @@ impl CpuInfo {
 
     /// Parse the CPU pool and build the list of off-line CPUs.
     fn get_cpu_info() -> NitroCliResult<Vec<u32>> {
-        let pool_file =
-            File::open("/sys/module/nitro_enclaves/parameters/ne_cpus").map_err(|e| {
-                new_nitro_cli_failure!(
-                    &format!("Failed to open CPU pool file: {}", e),
-                    NitroCliErrorEnum::FileOperationFailure
-                )
-            })?;
+        let pool_file = File::open(POOL_FILENAME).map_err(|e| {
+            new_nitro_cli_failure!(
+                &format!("Failed to open CPU pool file: {}", e),
+                NitroCliErrorEnum::FileOperationFailure
+            )
+            .add_info(vec![POOL_FILENAME, "Open"])
+        })?;
         let file_reader = BufReader::new(pool_file);
         let mut result: Vec<u32> = Vec::new();
 
@@ -173,6 +176,7 @@ impl CpuInfo {
                     &format!("Failed to read line from CPU pool file: {}", e),
                     NitroCliErrorEnum::FileOperationFailure
                 )
+                .add_info(vec![POOL_FILENAME, "Read"])
             })?;
             if line_str.trim().is_empty() {
                 continue;
--- a/src/enclave_proc/resource_manager.rs
+++ b/src/enclave_proc/resource_manager.rs
@@ -62,6 +62,9 @@ const NE_DEFAULT_MEMORY_REGION: u64 = 0;
 /// Magic number for Nitro Enclave IOCTL codes.
 const NE_MAGIC: u64 = 0xAE;
 
+/// Path corresponding to the Nitro Enclaves device file.
+const NE_DEV_FILEPATH: &str = "/dev/nitro_enclaves";
+
 /// IOCTL code for `NE_CREATE_VM`.
 pub const NE_CREATE_VM: u64 = nix::request_code_read!(NE_MAGIC, 0x20, size_of::<u64>()) as _;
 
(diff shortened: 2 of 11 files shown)


src/enclave_proc/resource_manager.rs
```diff
@@ -469,13 +470,26 @@ impl EnclaveHandle {
...
+            eif_file.as_raw_fd()
+        )));
+        let link_path;
+        let mut eif_path = "Unknown file";
+        if link_path_res.is_ok() {
+            link_path = link_path_res.unwrap();
+            eif_path = link_path
+                .to_str()
+                .unwrap_or("Invalid unicode EIF file name");
+        }
```
lexnv, line 484: Nit: we could wrap into a function 😄

lexnv (approved):

src/enclave_proc/resource_manager.rs
petreeftime, line 484: I don't think this is necessary. In such cases it would fail when you open the file, not here.

bercarug: I know, but the error related to invalid EIF name comes from the conversion from `std::fs::Path` to `str` implied by `link_path.to_str()` - in order to avoid `unwrap()`ping the result, I think the error case should be considered explicitly here.

alexggh:
Please remove the merge commit you have in your branch.

src/enclave_proc/resource_manager.rs
```diff
@@ -615,11 +630,24 @@ impl EnclaveHandle {
...
+                    .unwrap_or("Invalid unicode EIF file name"),
+            ),
+            Err(_) => String::from("Unknown file"),
+        };
         let eif_file = self.eif_file.as_mut().ok_or_else(|| {
             new_nitro_cli_failure!(
                 "Failed to get mutable reference to EIF file",
                 NitroCliErrorEnum::FileOperationFailure
             )
+            .add_info(vec![eif_path.as_str(), "Get EIF reference"])
```
petreeftime, line 650: This doesn't fail because of the EIF itself, I don't think, so reporting the file path is not needed here.

src/enclave_proc/resource_manager.rs
```diff
@@ -1007,13 +1035,26 @@ fn write_eif_to_regions(
...
+        Err(_) => String::from("Unknown file"),
+    };
     let file_size = eif_file
         .metadata()
         .map_err(|_| {
             new_nitro_cli_failure!(
                 "Failed to obtain EIF file metadata",
                 NitroCliErrorEnum::FileOperationFailure
             )
+            .add_info(vec![eif_path.as_str(), "Get metadata"])
```
petreeftime, line 1057: As far as I can tell, metadata will only fail here if the fd is bad, which is also not really related to the file itself, so I don't think we have to print out the file path here.

src/enclave_proc/resource_manager.rs
```diff
@@ -469,13 +470,26 @@ impl EnclaveHandle {
         debug_mode: bool,
     ) -> NitroCliResult<Self> {
         let requested_mem = memory_mib << 20;
+        let link_path = read_link(PathBuf::from(format!(
```
petreeftime, line 473: We're introducing extra error cases here which don't make sense I think. The metadata call can only likely only fail with EBADFD or some kernel error. read_link can fail in a thousand different ways though, and is not required. The failures are not necessarily related to the file itself, but tree traversal and other nonsence. I really don't think we need to do this here, as it's super error prone and doesn't improve logging IMO.

bercarug: Right, some of the File errors are not related directly to files stored on disk, but rather to the descriptor obtained after opening a file. In this case, `FileOperationFailures` would only correspond to actions failing because a file is not found or permissions don't allow for opening it etc.
Although the changes above deal with descriptors, I would not prefer to introduce an additional error code for something as `DescriptorOperationFailure`, since I don't think it really helps customers debug the problem themselves. Or should I consider adding this separate error code and use it above?

bercarug: I understand that `read_link` brings in extra complexity and possibilities for failures, but there's no other way of retrieving a file path starting from the fd, that I am aware of.

src/enclave_proc/resource_manager.rs
bercarug: Removed it in the latest revision.

src/enclave_proc/resource_manager.rs
bercarug: Removed it in the latest revision.

src/enclave_proc/resource_manager.rs
bercarug: Removed the file path from this section too. Indeed reporting the file path here would not help that much. And, since the error comes from an file descriptor operation for which `read_link` will most likely fail too.

src/enclave_proc/cpu_info.rs
```diff
@@ -157,13 +157,14 @@ impl CpuInfo {
...
     /// Parse the CPU pool and build the list of off-line CPUs.
     fn get_cpu_info() -> NitroCliResult<Vec<u32>> {
-        let pool_file =
-            File::open("/sys/module/nitro_enclaves/parameters/ne_cpus").map_err(|e| {
-                new_nitro_cli_failure!(
-                    &format!("Failed to open CPU pool file: {}", e),
-                    NitroCliErrorEnum::FileOperationFailure
-                )
-            })?;
+        let pool_filename = "/sys/module/nitro_enclaves/parameters/ne_cpus";
```
petreeftime, line 160: Define this as a const after imports.

src/enclave_proc/resource_manager.rs
```diff
@@ -504,6 +504,7 @@ impl EnclaveHandle {
                     &format!("Failed to open device file: {:?}", e),
                     NitroCliErrorEnum::FileOperationFailure
                 )
+                .add_info(vec!["/dev/nitro_enclaves", "Open"])
```
petreeftime, line 507: Define /dev/nitro_enclaves as const.

petreeftime (approved):
Approved with nits.

petreeftime (approved):

andraprs (approved):

Issue #196, opened 2020-11-06
nitro-cli run-enclave should enforce the allocated memory has space to unpack the initramfs


alexggh:
https://github.com/aws/aws-nitro-enclaves-cli/issues/188
https://github.com/aws/aws-nitro-enclaves-cli/issues/194

Seem to have the same root cause the allocated memory for the enclave is not enough for the kernel running inside the enclave to unpack the initramfs and user sees in the kernel logs:
- "Failed to unpack initramfs"

Followed by: 
- Could not open /env file: No such file or directory
- Could not open /env file: No such file or directory

The guideline to overcome this problem is to always have the enclave memory 4 times more than the size of Enclave image file(EIF). 

The nitro-cli should enforce this rule and provide an intuitive error when this happens.

bercarug:
Solved by https://github.com/aws/aws-nitro-enclaves-cli/pull/198

Issue #194, opened 2020-11-05
Segfault in init


jethrogb:
```
Could not open /env file: No such file or directory
Could not open /env file: No such file or directory
[    0.793543] init[1]: segfault at 0 ip 0000000000410e84 sp 00007fff27e27340 error 4 in init[400000+d4000]
```

The IP corresponds to fclose. You can't call fclose(NULL).

jethrogb:
This happens if you set `ENTRYPOINT` but not `ENV` or `CMD` in your Dockerfile.

alexggh:
> This happens if you set ENTRYPOINT but not ENV or CMD in your Dockerfile.

@jethrogb: I think you just helped us debug this: https://github.com/aws/aws-nitro-enclaves-cli/issues/188
Thank you

jethrogb:
> This happens if you set ENTRYPOINT but not ENV or CMD in your Dockerfile.

Not yet convinced this is the whole story.

jethrogb:
Are there supposed to be two EifSectionRamdisk sections in the Eif?

jethrogb:
I think there's some issue with initrd concatenation.

jethrogb:
This works:
```
# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0

FROM busybox

RUN truncate -s $((300*1024*1024)) /large_file

ENV HELLO="Hello from the enclave side!"

CMD ["/bin/sh", "-c", "while true; do echo \"$HELLO\"; sleep 10; done"]
```
This doesn't:
```
# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0

FROM busybox

RUN truncate -s $((350*1024*1024)) /large_file

ENV HELLO="Hello from the enclave side!"

CMD ["/bin/sh", "-c", "while true; do echo \"$HELLO\"; sleep 10; done"]
```

alexggh:
@jethrogb: You pointed to the exact error, the problem is the following you allocated to the enclave less memory than the linux kernel needs to unpack the ramfs. 

If you look at the kernel logs with the console command you would probably first see "Failed to unpack initiramfs"

As of now the guideline is that the enclave should have allocated at least 4 times more memory than the size of the EIF.

jethrogb:
Indeed it says:
```
[...]

Issue #188, opened 2020-11-03
env issue with python docker image


petreeftime:
Based on https://levelup.gitconnected.com/running-python-app-on-aws-nitro-enclaves-56024667b684, the docker python image does not start under an enclave. Investigate why this is happening:

> For Python app development, using Python base image would be an obvious step. But for some reason, the python image is not running properly in the Nitro Enclave.
> No matter I use the standard image, -slim image or -alpine image, even my Dockerfile is just running a simple echo command, an error Could not open /env file: No such file or directory will come up.

bercarug:
Solved by https://github.com/aws/aws-nitro-enclaves-cli/pull/198

PR #198, opened 2020-11-06
cli: Ensure enclave is started with enough memory

bercarug:
Currently, before starting an enclave, the user provided
enclave memory is checked to be at least equal to the EIF
size.
However, this constraint is not strong enough and in order
to ensure there is enough memory to unpack the initramfs,
this commit introduces a lower limit on the enclave memory:
4 times the size of the EIF file.

*Issue #, if available: #188, #194, #196*

### DIFF
--- a/src/enclave_proc/resource_manager.rs
+++ b/src/enclave_proc/resource_manager.rs
@@ -37,6 +37,7 @@ use crate::enclave_proc::cpu_info::EnclaveCpuConfig;
 use crate::enclave_proc::utils::get_run_enclaves_info;
 use crate::enclave_proc::utils::{GiB, MiB};
 use crate::new_nitro_cli_failure;
+use crate::utils::ceil_div;
 
 /// CamelCase alias for the bindgen generated driver struct (ne_enclave_start_info).
 pub type EnclaveStartInfo = ne_enclave_start_info;
@@ -52,6 +53,9 @@ type UnpackedHandle = (u64, u64, u64, Vec<u32>, u64, u64, EnclaveState);
 
 /// The bit indicating if an enclave has been launched in debug mode.
 pub const NE_ENCLAVE_DEBUG_MODE: u64 = 0x1;
+
+/// Constant number used for computing the lower memory limit.
+const ENCLAVE_MEMORY_EIF_SIZE_RATIO: u64 = 4;
 
 /// Enclave Image Format (EIF) flag.
 const NE_EIF_IMAGE: u64 = 0x01;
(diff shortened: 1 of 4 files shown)


src/enclave_proc/resource_manager.rs
```diff
@@ -53,6 +53,9 @@ type UnpackedHandle = (u64, u64, u64, Vec<u32>, u64, u64, EnclaveState);
 /// The bit indicating if an enclave has been launched in debug mode.
 pub const NE_ENCLAVE_DEBUG_MODE: u64 = 0x1;
 
+/// Constant number used for computing the lower memory limit.
+const NE_EIF_MEM_THRESHOLD: u64 = 4;
```
alexggh, line 57: ENCLAVE_MEMORY_EIF_SIZE_RATIO would be more suggestive.

src/enclave_proc/resource_manager.rs
```diff
@@ -481,19 +484,18 @@ impl EnclaveHandle {
                 )
             })?
             .len();
-        if eif_size > requested_mem {
+        if NE_EIF_MEM_THRESHOLD * eif_size > requested_mem {
             return Err(new_nitro_cli_failure!(
                 &format!(
                     "At least {} MiB must be allocated (which is the EIF file size)",
```
alexggh, line 490: The message does not seem to be up to date anymore.

src/enclave_proc/resource_manager.rs
```diff
@@ -1087,6 +1089,19 @@ fn release_enclave_descriptor(enc_fd: RawFd) -> NitroCliResult<()> {
     Ok(())
 }
 
+/// Computes the ceil of `lhs / rhs`.
```
alexggh, line 1092: Can we place this function in utils ?

src/enclave_proc/resource_manager.rs
bercarug: Renamed in the latest revision.

src/enclave_proc/resource_manager.rs
bercarug: Right, updated in the latest revision.

src/enclave_proc/resource_manager.rs
bercarug: Placed it for the moment in utils.rs, however at the moment that file contains things mainly related to the console functionality. Maybe in a future update we will split the file in a proper `utils.rs` and a `console.rs`?

andraprs (approved):

alexggh (approved):

Issue #193, opened 2020-11-05
nitro-cli console should report if debug-flag is not present


alexggh:
In the case where debug-mode is not passed when running the enclave the  nitro-cli console won't be able to connect to it. 

However,  the cli has all the information it needs to report a proper error if that flag is not present instead of reporting that it could not connect to the enclave console which could happen from multiple reasons.

PR #203, opened 2020-11-19
nitro-cli: Check debug flag is present for enclave console access

andraprs:
The enclave needs to be started in debug mode to get its console output.
Check the enclave flags and early exit if the debug mode flag is not
set when the enclave is created.

*Issue #193*

*Description of changes:*

### DIFF
--- a/src/common/mod.rs
+++ b/src/common/mod.rs
@@ -191,6 +191,8 @@ pub enum EnclaveProcessCommandType {
     Describe,
     /// Request an enclave's CID (sent by the CLI).
     GetEnclaveCID,
+    /// Request an enclave's flags (sent by the CLI).
+    GetEnclaveFlags,
     /// Notify the socket connection listener to shut down (sent by the enclave process to itself).
     ConnectionListenerStop,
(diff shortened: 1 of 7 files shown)


src/common/mod.rs
```diff
@@ -189,8 +189,8 @@ pub enum EnclaveProcessCommandType {
     TerminateComplete,
     /// Describe an enclave (broadcast by the CLI).
     Describe,
-    /// Request an enclave's CID (sent by the CLI).
-    GetEnclaveCID,
+    /// Request an enclave's CID and flags (sent by the CLI).
+    GetEnclaveCIDAndFlags,
```
petreeftime, line 193: I would prefer to have a GetEnclaveCID and GetEnclaveFlags separately, or drop both and use Describe instead.

lexnv (dismissed):

petreeftime (changes requested):

bveaws (dismissed):

petreeftime (approved):

alexggh (approved):

PR #208, opened 2020-11-27
packaging: Prevent file erasure on package upgrade

bercarug:
Currently, plain RPM package upgrades leave the nitro-cli
into an unusable state, since the post-uninstall scriptlet
removes necessary files and directories.

Checking the package operation in both pre and post
uninstall is necessary in order to successfully run
nitro-cli commands after performing an install operation
(e.g. `sudo yum update aws-nitro-enclaves-cil`).
The meaning of the parameter values are described here:
https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/.

Instead of relying on the user to manually update his
already installed v1.0-5 of the Nitro CLI package, this
commit introduces a separate `%posttrans` scriptlet
that currently performs the same `%post` steps as before,
in order to reconfigure the environment after the old
package version uninstall.
    
This commit also updates the package version to 1.1-1.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nitro-cli"
-version = "1.0.8"
+version = "1.0.9"
 authors = ["Alexandru Gheorghe <aggh@amazon.com>"]
 edition = "2018"
 
--- a/SPECS/aws-nitro-enclaves-cli.spec
+++ b/SPECS/aws-nitro-enclaves-cli.spec
@@ -17,8 +17,8 @@
 
 Summary:    AWS Nitro Enclaves tools for managing enclaves
 Name:       aws-nitro-enclaves-cli
-Version:    1.0
-Release:    8%{?dist}
+Version:    1.0.9
+Release:    0%{?dist}
 
 License:    Apache 2.0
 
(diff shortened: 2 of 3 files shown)


SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -130,18 +130,24 @@ echo -e "
...
 
 %preun
-# Uninstall services
-%systemd_preun nitro-enclaves-vsock-proxy.service
-%systemd_preun nitro-enclaves-allocator.service
+if [ $1 -ne 1 ]; then
+    # Any operation except for package upgrade
+    # Uninstall services
+    %systemd_preun nitro-enclaves-vsock-proxy.service
+    %systemd_preun nitro-enclaves-allocator.service
```
bveaws, line 137: Aren't these macros already handling the checks themselves ?

bercarug: Indeed, this is the case. Will remove the redundant check.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "nitro-cli"
-version = "1.0.8"
+version = "1.1.1"
```
petreeftime, line 3: why 1.1.1 and not 1.1.0? Also, why is this a version change, I don't see anything in the changelog that would indicate any significant enough functionality change.

SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -136,12 +136,34 @@ echo -e "
 
 
 %postun
-# Remove any directory which was created by the driver as well as unload the driver
-rm -f /usr/lib/modules-load.d/nitro_enclaves.conf
-rm -f /usr/lib/udev/rules.d/99-nitro_enclaves.rules
-rm -f /usr/lib/tmpfiles.d/nitro_enclaves.conf
-rm -rf %{ne_run_dir}
-rm -rf %{ne_log_dir}
+if [ $1 -ne 1 ]; then
```
petreeftime, line 139: I think also on install you wouldn't want to remove all those file which people might have installed by hand.

Cargo.toml
bercarug: Right, revision could be 0. The version change was more of a convention to mark that versions before 1.1 (namely 1.0-5 which is currently available to customers on AL2 instances) misbehave on `yum update`s. Versions from 1.1 fix this update problem.

SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -17,8 +17,8 @@
 
 Summary:    AWS Nitro Enclaves tools for managing enclaves
 Name:       aws-nitro-enclaves-cli
-Version:    1.0
-Release:    8%{?dist}
+Version:    1.1
```
petreeftime, line 20: Version here should actually read 1.1.0 instead of 1.1; or actually 1.0.9 or something. The release number is a separate thing and it should not be related to the version number of the project itself.

SPECS/aws-nitro-enclaves-cli.spec
bercarug: I don't think the `%postun` scriptlet is run on package installation.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "nitro-cli"
-version = "1.0.8"
+version = "1.0.9"
```
andraprs, line 3: Should add also the update in Cargo.lock (https://github.com/aws/aws-nitro-enclaves-cli/blob/main/Cargo.lock#L923).

SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -17,8 +17,8 @@
 
 Summary:    AWS Nitro Enclaves tools for managing enclaves
 Name:       aws-nitro-enclaves-cli
-Version:    1.0
-Release:    8%{?dist}
+Version:    1.0.9
```
andraprs, line 20: Should '9' be in the Release field?

andraprs (approved):

Cargo.toml
bercarug: Right, will update it in the next revision.

petreeftime (approved):

SPECS/aws-nitro-enclaves-cli.spec
bercarug: From offline discussions we decided to switch to a version format such as `1.0.9-0`.

andraprs: Ok, thanks for info.

andraprs (approved):

PR #218, opened 2021-01-25
build: Make the build on ARM functional

laurstefan:
This PR makes the nitro-cli build on ARM.

Signed-off-by: Laurentiu Stefan <stefala@amazon.com>

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,10 +1,50 @@
 extern crate bindgen;
 
+use std::env;
+use std::io;
 use std::path::PathBuf;
 use std::process::Command;
 
 const HEADER_PATH: &str = "include/uapi/linux/nitro_enclaves.h";
 const OUT_FILE: &str = "/driver_structs.rs";
+
+// Generate the sysroot path based on the architecture the program is built.
+// This is needed for the compiler to locate the architecture dependent
+// header files.
+fn get_gcc_sysroot() -> String {
+    // Initialize the output because it might remain uninitialized with the
+    // current control flow
+    let mut output = Err(io::Error::from(io::ErrorKind::NotFound));
+
+    // Use the `CC` environment variable if existing
+    // which should give us directly the compiler's name
+    let cc = env::var("CC");
+    if let Ok(cc) = cc {
+        output = Command::new(cc).arg("--print-sysroot").output();
+    }
+
+    if let Err(_) = output {
+        // Otherwise, try to guess the compiler's name by the name
+        // $TARGET-gcc, where $TARGET represents the target architecture
+        // that was specified for build
+        let target = env::var("TARGET").unwrap();
+        output = Command::new(format!("{}-gcc", target))
+            .arg("--print-sysroot")
+            .output();
+        if let Err(_) = output {
+            // If that did not work as well, use directly `gcc`
+            output = Command::new("gcc").arg("--print-sysroot").output();
+        }
+    }
+
+    // At this point we should have a valid command output
+    // that gives us the sysroot path
+    let output = output.expect("no compiler was found for the current target");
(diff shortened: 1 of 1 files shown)


petreeftime:
Can you amend the commit text to explain why this is required and what the status before and after the change is?

build.rs
```diff
@@ -1,11 +1,24 @@
 extern crate bindgen;
 
+use std::env;
 use std::path::PathBuf;
 use std::process::Command;
 
 const HEADER_PATH: &str = "include/uapi/linux/nitro_enclaves.h";
 const OUT_FILE: &str = "/driver_structs.rs";
 
+fn get_gcc_sysroot() -> String {
```
lucianperju, line 10: Some comments would be really useful. We should be able why this is needed anytime we open this file.

laurstefan: Added and amended.

laurstefan:
> Can you amend the commit text to explain why this is required and what the status before and after the change is?

I've added some information in the commit message. Let me know if it is more clear now.

build.rs
```diff
@@ -1,11 +1,50 @@
...
+// This is needed for the compiler to locate the architecture dependent
+// header files.
+fn get_gcc_sysroot() -> String {
+    // Use the `CC` environment variable if existing
+    // which should give us directly the compiler's name
+    let cc = env::var("CC");
+    if let Ok(cc) = cc {
+        let output = Command::new(cc).arg("--print-sysroot").output();
+        if let Ok(output) = output {
+            let output = String::from_utf8_lossy(&output.stdout);
```
alexggh, line 20: Could you refactor the code a bit, so that we don't duplicate this part.
Something like:
First detect the compiler then parse the output, something like you already did for the last two code paths.

laurstefan: Done.

build.rs
```diff
@@ -1,11 +1,55 @@
...
+    // which should give us directly the compiler's name
+    let cc = env::var("CC");
+    if let Ok(cc) = cc {
+        output = Command::new(cc).arg("--print-sysroot").output();
+        if let Ok(_) = output {
+            cc_found = true;
+        }
+    }
+
+    if !cc_found {
```
alexggh, line 30: why isn't this on the else branch ?

build.rs
```diff
@@ -1,11 +1,55 @@
...
+            .output();
+        if let Err(_) = output {
+            // If that did not work as well, use directly `gcc`
+            output = Command::new("gcc").arg("--print-sysroot").output();
+        }
+    }
+
+    // At this point we should have a valid command output
+    // that gives us the sysroot path
+    let output = output.unwrap();
```
alexggh, line 46: it should be an expect with a helpfull error message so that the user knows what to do :D.

build.rs
laurstefan: Because there are two situations that can fail:
1. Environment variable "CC" is not set
2. The command "$CC --print-sysroot" fails for any reason (such as $CC is not set correctly)

If I just add the code on the else statement of the first if, the second case would not be covered.

build.rs
laurstefan: That could be helpful indeed. Will change.

build.rs
alexggh: Rigght but it could be:
if let Ok(cc) = cc {
....
}
if let Err(_) = output {
....
}
as you did bellow

laurstefan: Yep, here I agree :D Will change it.

alexggh (dismissed):

petreeftime (dismissed):

petreeftime (approved):

lexnv (approved):

PR #219, opened 2021-01-26
eif: Add EIF header architecture flag

alcioa:
When building an EIF image, inspect the source docker
image and based on its architecture (i.e. amd64 vs aarch64)
set the flag accordingly in the EIF header flags.

This flag shall be cleared for amd64 and toggled for aarch64.

Signed-off-by: Alexandru Ciobotaru <alcioa@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,6 +391,7 @@ version = "0.1.0"
 dependencies = [
  "base64",
  "clap",
+ "eif_defs",
  "eif_utils",
  "log",
  "serde",
--- a/eif_defs/src/lib.rs
+++ b/eif_defs/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2019-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 #![deny(warnings)]
@@ -18,6 +18,10 @@ use std::mem::size_of;
 
 pub const EIF_MAGIC: [u8; 4] = [46, 101, 105, 102]; // .eif in ascii
 pub const MAX_NUM_SECTIONS: usize = 32;
+/// EIF Header Flags
+/// bits 1  : Architecture - toggled for aarch64, cleared for x86_64
+/// bits 2-8: Unused
+pub const EIF_HDR_ARCH_ARM64: u16 = 0x1;
 
 /// Current EIF version to be incremented every time we change the format
 /// of this structures, we assume changes are backwards compatible.
@@ -31,6 +35,7 @@ pub struct EifHeader {
     pub magic: [u8; 4],
     /// EIF version
     pub version: u16,
+    /// EIF header flags
     pub flags: u16,
     /// Default enclave memory used to boot this eif file
     pub default_mem: u64,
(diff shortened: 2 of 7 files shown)


enclave_build/src/lib.rs
```diff
@@ -181,11 +183,22 @@ impl<'a> Docker2Eif<'a> {
             return Err(Docker2EifError::LinuxkitExecError);
         }
 
+        let arch = self.docker.architecture().map_err(|e| {
+            eprintln!("Docker error: {:?}", e);
+            Docker2EifError::DockerError
+        })?;
+
+        let flags: u16 = if arch == "arm64" {
```
lucianperju, line 191: arm64 should be a const defined somewhere

enclave_build/src/docker.rs
```diff
@@ -14,6 +14,10 @@ use tempfile::NamedTempFile;
 use tokio::prelude::{Future, Stream};
 use url::Url;
 
+/// Docker inspect architecture constants
+pub const DOCKER_ARCH_ARM64: &str = "arm64";
+pub const DOCKER_ARCH_AMD64: &str = "amd64";
```
petreeftime, line 19: I'd go for the more commonly used aarch64 and x86_64 instead, especially since it's difficult to see the difference between arm64 and amd64.

alcioa: That was the intention but `docker inspect` returns these constants.

eif_defs/src/lib.rs
```diff
@@ -18,6 +18,10 @@ use std::mem::size_of;
 
 pub const EIF_MAGIC: [u8; 4] = [46, 101, 105, 102]; // .eif in ascii
 pub const MAX_NUM_SECTIONS: usize = 32;
+/// EIF Header Flags
+/// bits 1  : Architecture - toggled for aarch64, cleared for x86_64
+/// bits 2-8: Unused
+pub const EIF_HDR_ARCH_ARM64: u16 = 0x1;
```
petreeftime, line 24: What if we ever need add RISC-V or something else? We should have a separate arch field.

petreeftime: Given that we might not want to add another version of EIF right now, I think this is fine.

Cargo.lock
```diff
@@ -391,6 +391,7 @@ version = "0.1.0"
 dependencies = [
  "base64",
  "clap",
+ "eif_defs",
```
alexggh, line 394: why is this needed now ?

alcioa: Because `enclave_build` now depends on `eif_defs` and `Cargo.lock` required update.

enclave_build/src/lib.rs
```diff
@@ -181,11 +184,23 @@ impl<'a> Docker2Eif<'a> {
             return Err(Docker2EifError::LinuxkitExecError);
         }
 
+        let arch = self.docker.architecture().map_err(|e| {
+            eprintln!("Docker error: {:?}", e);
+            Docker2EifError::DockerError
+        })?;
+
+        let flags = match arch.as_str() {
```
alexggh, line 192: I think the right abstraction is to pass the arch string to EifBuilder and let that object take the decision.
And this way you avoid making enclave_build dependent on eif_defs.

alcioa: This would require switching `eif_utils` dependent on `enclave_build/docker.rs`. Let's pass the flag instead of the arch string from here for now.

bercarug (approved):

lucianperju (approved):

Issue #220, opened 2021-01-28
`make nitro-cli` fails on Ubuntu 18.04


dreemkiller:
When following the build instructions, I get an error.
I've isolated it to the .build_container target, and the error I get when I run `make nitro-cli`:

```
Step 23/27 : RUN  source $HOME/.cargo/env &&    cargo install cargo-audit --version ^0.12
 ---> Running in 02e0cbf2c49d
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-audit v0.12.1
  Installing cargo-audit v0.12.1
 Downloading crates ...
  Downloaded thiserror v1.0.23
  Downloaded rustsec v0.20.1
  Downloaded gumdrop v0.7.0
  Downloaded lazy_static v1.4.0
  Downloaded serde v1.0.123
  Downloaded home v0.5.3
  Downloaded smol_str v0.1.16
  Downloaded abscissa_core v0.5.2
  Downloaded serde_json v1.0.61
  Downloaded gumdrop_derive v0.7.0
  Downloaded chrono v0.4.19
  Downloaded thiserror-impl v1.0.23
  Downloaded serde_derive v1.0.123
  Downloaded toml v0.5.8
  Downloaded semver v0.9.0
  Downloaded cvss v1.0.1
  Downloaded platforms v0.2.1
  Downloaded semver-parser v0.9.0
  Downloaded cargo-lock v4.0.1
  Downloaded crates-index v0.14.3
  Downloaded git2 v0.13.17
  Downloaded num-traits v0.2.14
  Downloaded once_cell v1.5.2
  Downloaded canonical-path v2.0.2
  Downloaded signal-hook v0.1.17
  Downloaded secrecy v0.6.0
  Downloaded backtrace v0.3.56
  Downloaded libc v0.2.83
  Downloaded tracing v0.1.22
  Downloaded regex v1.4.3
  Downloaded semver-parser v0.7.0
  Downloaded ryu v1.0.5
  Downloaded termcolor v1.1.2
  Downloaded time v0.1.44
  Downloaded quote v1.0.8
  Downloaded proc-macro2 v1.0.24
  Downloaded syn v1.0.60
  Downloaded tracing-log v0.1.1
  Downloaded color-backtrace v0.3.0
  Downloaded itoa v0.4.7
  Downloaded tracing-subscriber v0.1.6
  Downloaded num-integer v0.1.44
  Downloaded wait-timeout v0.2.0
[...]

PR #221, opened 2021-01-28
Changed the RUST_VERSION to 1.49 to solve issue #220

dreemkiller:
*Issue #, if available:*
This is **a** fix for #220. I've no idea if is the appropriate way to solve this for the project.

*Description of changes:*
In file tools/Dockerfile1804, changed the line:
```
ENV RUST_VERSION 1.41.1
```
to:
```
ENV RUST_VERSION 1.49
```
This changes the version of rust that is compiling the code.

### DIFF
--- a/tools/Dockerfile1804
+++ b/tools/Dockerfile1804
@@ -28,7 +28,7 @@ RUN unzip /openssl_src/${OPENSSL_VERSION}.zip -d /openssl_src
 RUN cd /openssl_src/openssl-${OPENSSL_VERSION} && CC=musl-gcc CFLAGS=-fPIC ./Configure --prefix=/musl_openssl --openssldir=/musl_openssl no-shared no-engine no-afalgeng linux-x86_64 -DOPENSSL_NO_SECURE_MEMORY && make && make install
 
 # Setup the right rust ver
-ENV RUST_VERSION=1.49
+ENV RUST_VERSION=1.49.0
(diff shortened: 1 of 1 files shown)


petreeftime (approved):
It seems that this feature was included in 1.42, so a more conservative approach would be to update to that instead. I don't see a reason why not to use 1.49 if the project compiles with it though, as long as we can test that it builds with something older as well, but that would have to be part of another PR. Any other thoughts on this?

petreeftime:
It seems that the version should be listed as `1.49.0`, otherwise it wouldn't build. Likely the container could do with a newer rustup version, but could you please list the version exactly?

dreemkiller:
I've updated it to be specific at 1.49.0, and it does work. Again, it is up to you to decide if you actually want to upgrade your toolchain this way.

alcioa:
Hi, the issue is that running `make nitro-format` and `make nitro-clippy` shall fail with 1.49.0 due to the Rust toolchain upgrade.

dreemkiller:
As mentioned in the filed issue, then this may not be the way you want to move forward. The other alternative is to be more specific in the version of `cargo-audit` that you specify.

This pull request solves the problem for me. If you'd like to take the alternate approach to solving the problem, please feel free to reject this PR and solve it in the way most appropriate for the project.

alcioa:
@dreemkiller we're closing this as #224 with this version is now in main branch.

PR #222, opened 2021-02-01
build: split build and blobs by architecture

alcioa:
Split the blobs and examples into their own directories
by architecture and add the aarch64 Dockerfile.

Builds on aarch64 machine. No cross build support/testing.

Signed-off-by: Alexandru Ciobotaru <alcioa@amazon.com>

*Issue #, if available:*

*Description of changes:*


Makefile
```diff
@@ -8,6 +8,11 @@
 
 .DEFAULT_GOAL := nitro-cli
 
+ARCH_x86_64      = x86_64
```
alexggh, line 11: pull requests checks are failing, could you have a look to see what is going on ?

alcioa: Sorry, forgot to mention yesterday. The bump to Rust 1.49.0 bump shall cause `make nitro-clippy` and `make nitro-format` to fail. We shall fix the 1.49.0 issues in a separate PR and this PR should pass after I'll rebase and repush.

alexggh (approved):
Looks good to me.
Why did we change the sample blobs ?

alcioa:
> Looks good to me.
> Why did we change the sample blobs ?

The sample blobs were symbolic links to the root directory blobs and they had to be reconstructed with the new per-arch directory layout.

andraprs (dismissed):
May need also an update to default blobs path, based on arch options. Should  double-check the RPM spec, if all good with blobs setup at the given path.

https://github.com/aws/aws-nitro-enclaves-cli/blob/main/src/lib.rs#L41

alcioa:
> May need also an update to default blobs path, based on arch options. Should double-check the RPM spec, if all good with blobs setup at the given path.
> 
> https://github.com/aws/aws-nitro-enclaves-cli/blob/main/src/lib.rs#L41

I initially took a look at that path. That's the same for each architecture, i.e. where we copy these blobs to that path. So for both archs we now copy them from the `blobs/<arch>/` to `/usr/share/nitro-enclaves`. Had a chat cu bercarug@ so https://github.com/aws/aws-nitro-enclaves-cli/blob/main/SPECS/aws-nitro-enclaves-cli.spec#L186 is the post-install `/usr/share/nitro-enclaves/blobs` path which is right now common for both archs. Of course, we do not have ARM RPM spec right now. Thanks.

andraprs:
> > May need also an update to default blobs path, based on arch options. Should double-check the RPM spec, if all good with blobs setup at the given path.
> > https://github.com/aws/aws-nitro-enclaves-cli/blob/main/src/lib.rs#L41
> 
> I initially took a look at that path. That's the same for each architecture, i.e. where we copy these blobs to that path. So for both archs we now copy them from the `blobs/<arch>/` to `/usr/share/nitro-enclaves`. Had a chat cu bercarug@ so https://github.com/aws/aws-nitro-enclaves-cli/blob/main/SPECS/aws-nitro-enclaves-cli.spec#L186 is the post-install `/usr/share/nitro-enclaves/blobs` path which is right now common for both archs. Of course, we do not have ARM RPM spec right now. Thanks.

Ok, thanks for checking it out.

andraprs (approved):

alexggh (approved):

PR #223, opened 2021-02-02
SPEC: use a trigger script instead of %posttrans

bercarug:
Replaces the post transaction scriptlet with a
trigger script which re-initializes the resources
only then the post-installation phase of package
v1.0-5 occurrs.
This indeed helps when the package is updated from
version 1.0-5 to whatever latest version might
be available at the update time (not necessarily the
consecutive release after 1.0-5).

### DIFF
--- a/SPECS/aws-nitro-enclaves-cli.spec
+++ b/SPECS/aws-nitro-enclaves-cli.spec
@@ -6,9 +6,9 @@
 %define ne_data_dir %{_datadir}/%{ne_name}
 %define ne_include_dir %{_includedir}/%{ne_name}
 %define ne_sysconf_dir %{_sysconfdir}/%{ne_name}
-%define ne_log_dir /var/log/%{ne_name}
+%define ne_log_dir %{_localstatedir}/log/%{ne_name}
 %define ne_log_file %{ne_name}.log
-%define ne_run_dir /run/%{ne_name}
+%define ne_run_dir %{_rundir}/%{ne_name}
 
 %define _src_dir %{_builddir}/%{name}-%{version}
 %define _licenses_filename THIRD_PARTY_LICENSES
@@ -17,7 +17,7 @@
 
 Summary:    AWS Nitro Enclaves tools for managing enclaves
 Name:       aws-nitro-enclaves-cli
-Version:    1.0.9
+Version:    1.0.10
 Release:    0%{?dist}
 
 License:    Apache 2.0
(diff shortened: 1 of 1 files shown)


SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -6,9 +6,9 @@
 %define ne_data_dir %{_datadir}/%{ne_name}
 %define ne_include_dir %{_includedir}/%{ne_name}
 %define ne_sysconf_dir %{_sysconfdir}/%{ne_name}
-%define ne_log_dir /var/log/%{ne_name}
+%define ne_log_dir %{_localstatedir}/log/%{ne_name}
```
lexnv, line 9: I m wondering about the failing status of the testing. Could this be the offending change?

bercarug: Most likely not, the SPEC file is not exercised by the integration tests - looking into what is causing this.

petreeftime (dismissed):

SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -188,6 +188,10 @@ udevadm trigger -y nitro_enclaves
 %{ne_include_dir}/*
 
 %changelog
+* Tue Feb 02 2021 Gabriel Bercaru <bercarug@amazon.com> - 1.0.10-0
+- Removed the %posttrans scriptlet and delegated the task of
+  re-performing resources initialization to a trigger script
+  which runs only then uninstalling v1.0 or the package
```
andraprs, line 194: nit: s/then/when/

andraprs, line 194: Would the trigger be run also during uninstall (at all, not during package update)? Should not be needed then. Just clarify a bit the text here "... or the package".

andraprs (approved):

SPECS/aws-nitro-enclaves-cli.spec
bercarug: Done in the next revision

bercarug: Yes, it is run at plain uninstall, but in order to happen, the user should have both this version (which uses this spec file version, for example `aws-nitro-enclaves-cli-1.0.10-0`) and the previous package version (`aws-nitro-enclaves-cli` v1.0). If this is possible than indeed the trigger script should be updated in order to perform only on package update.

andraprs: Thanks for info. Right, if that's needed, a check should be for the package action, if update or removal at all.

bercarug: I tested and saw that the package operation during the execution of the trigger script is `2`, so I added the corresponding check, to be on the safe side (even tough I couldn't get the trigger script to run when doing a simple package removal).

andraprs (approved):

andraprs (approved):

axlprv (approved):

Issue #239, opened 2021-04-05
`nitro-cli build-enclave` failing with DockerError


gkelly:
If I use the `nitro-cli` that's packaged with the `amazon-linux-extras install aws-nitro-enclaves-cli` I can run `build-enclave` without issue (`nitro-cli version` reports `Nitro CLI 1.0.9`). But building the current `main` branch on the same EC2 instance that can successfully build with the packaged version (using the same arguments) results in:

```
./target/debug/nitro-cli build-enclave --docker-uri hello-enclave:1.0 --output-file what.eif
Start building the Enclave Image...
Docker error: PullError
[ E50 ] Docker image pull error. Such error appears when trying to build an EIF file, but pulling the corresponding docker image fails. In this case, the error backtrace provides detailed informatino on the failure reason.

For more details, please visit https://docs.aws.amazon.com/enclaves/latest/user/cli-errors.html#E50

If you open a support ticket, please provide the error log found at "/var/log/nitro_enclaves/err2021-04-05T21:34:26.317152048+00:00.log"
```

And the `/var/log/nitro_enclaves/err2021-04-05T21:34:26.317152048+00:00.log` file contains:
```
 Action: Build Enclave
  Subactions:
    Failed to build enclave
    Failed to build EIF from docker
    Failed to pull docker image: DockerError
  Root error file: src/lib.rs
  Root error line: 145
  Build commit: 4d51695
```

gkelly:
I don't think there should be a pull involved in this `build-enclave`, as the image is fully local.

gkelly:
Adding to this, if I use a `--docker-uri` that _can_ be pulled then the build succeeds:

```
[ec2-user@ip-10-1-1-100 aws-nitro-enclaves-cli]$ ./target/debug/nitro-cli build-enclave --docker-uri hello-world:latest --output-file what.eif
Start building the Enclave Image...
Enclave Image successfully created.
{
  "Measurements": {
    "HashAlgorithm": "Sha384 { ... }",
    "PCR0": "71e7680ae9c52032a98293899d93315310ff4787fc9a312027407fed94120b544c08fe440cc6371ad488109e41ab05db",
    "PCR1": "c35e620586e91ed40ca5ce360eedf77ba673719135951e293121cb3931220b00f87b5a15e94e25c01fecd08fc9139342",
[...]

PR #240, opened 2021-04-06
nitro-cli: Fix Docker pull image logic

andraprs:
Check if the Docker image used for building the enclave image is locally
available, before trying to pull a remote Docker image. Early exit if
the Docker image is locally available.

Update the tests to use the locally built Docker image of the command
executer, instead of the remote one, to cover this test case.

*Issue #239*

*Description of changes:*

### DIFF
--- a/enclave_build/src/docker.rs
+++ b/enclave_build/src/docker.rs
@@ -174,6 +174,20 @@ impl DockerUtil {
     /// Pull the image, with the tag provided in constructor, from the Docker registry
     pub fn pull_image(&self) -> Result<(), DockerError> {
         let act = async {
+            // Check if the Docker image is locally available.
+            // If available, early exit.
+            if self
+                .docker
+                .images()
+                .get(&self.docker_image)
+                .inspect()
+                .await
+                .is_ok()
+            {
+                eprintln!("Using the locally available Docker image...");
+                return Ok(());
+            }
+
             let mut pull_options_builder = PullOptions::builder();
             pull_options_builder.image(&self.docker_image);
 
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -25,6 +25,9 @@ function clean_up_and_exit() {
 	[ "$(lsmod | grep -cw nitro_enclaves)" -eq 0 ] || rmmod nitro_enclaves || register_test_fail
 	make clean
 	rm -rf test_images
+	# Cleanup pulled images during testing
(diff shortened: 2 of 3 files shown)


enclave_build/src/docker.rs
```diff
@@ -205,6 +205,15 @@ impl DockerUtil {
                             }
                         }
                         Err(e) => {
+                            // Check if the Docker image is locally available and not remotely.
+                            // If not locally available return error.
+                            if let Ok(_image) =
```
gkelly, line 210: I think the behavior here might be surprising to users. Please correct me, but I believe the logic is now:

1. Attempt to pull the image.
2. If the pull fails, check if the image exists locally.

This now requires that if a user build a docker image locally it must not have a name that can be pulled, or else the EIF build _will_ pull it, overwriting their local image.

I believe the unsurprising behavior is to check if the image is locally available _first_, and if it's not _then_ attempt to pull it.

andraprs: Correct, that's how logic currently works. I was thinking of this case, just that the previous logic, before the crate update, was pulling the image anyway, but not exposing the possible error when trying to fetch from the remote repo.

Anyhow, that's a fair point, I can update the logic to check for local images first. Thanks for bringing it up.

gkelly (dismissed):
Thanks @andraprs, I think this is a great solution.

enclave_build/src/docker.rs
```diff
@@ -174,6 +174,13 @@ impl DockerUtil {
     /// Pull the image, with the tag provided in constructor, from the Docker registry
     pub fn pull_image(&self) -> Result<(), DockerError> {
         let act = async {
+            // Check if the Docker image is locally available.
+            // If available, early exit.
+            if let Ok(_image) = self.docker.images().get(&self.docker_image).inspect().await {
```
bercarug, line 179: Nit: will the `.await` at the end force the evaluation of the yielded `Future` object, producing a `Result`? If this is indeed the case and the outcome is a result, could we call `.is_ok()` on it?

Something like `if self.docker.images().get(&self.docker_image).inspect().await.is_ok() { ... }`, since the `_image` wrapped inside the `Ok` is not used?

bercarug (approved):

andraprs:
> Thanks @andraprs, I think this is a great solution.

You are welcome.

enclave_build/src/docker.rs
andraprs: Yup, that will work, thanks for the note. I updated the logic.

bercarug (approved):

PR #249, opened 2021-04-22
vsock_proxy: Update nitro-enclaves-vsock-proxy service to use IMDSv2

andraprs:
The start command for the nitro-enclaves-vsock-proxy service uses IMDS
to retrieve the region corresponding to the running instance.

Update the start command to use IMDSv2.

*Issue #, if available:*

*Description of changes:*

Documentation for IMDSv2: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

### DIFF
--- a/vsock_proxy/service/nitro-enclaves-vsock-proxy.service
+++ b/vsock_proxy/service/nitro-enclaves-vsock-proxy.service
@@ -8,7 +8,8 @@ Type=simple
 StandardOutput=journal
 StandardError=journal
 SyslogIdentifier=vsock-proxy
-ExecStart=/bin/bash -ce "REGION=$(curl --silent http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .region) ; \
+ExecStart=/bin/bash -ce "TOKEN=$(curl --silent -X PUT \"http://169.254.169.254/latest/api/token\" -H \"X-aws-ec2-metadata-token-ttl-seconds: 21600\") ; \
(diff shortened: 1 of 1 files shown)


vsock_proxy/service/nitro-enclaves-vsock-proxy.service
```diff
@@ -8,7 +8,8 @@ Type=simple
 StandardOutput=journal
 StandardError=journal
 SyslogIdentifier=vsock-proxy
-ExecStart=/bin/bash -ce "REGION=$(curl --silent http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .region) ; \
+ExecStart=/bin/bash -ce "TOKEN=$(curl --silent -X PUT \"http://169.254.169.254/latest/api/token\" -H \"X-aws-ec2-metadata-token-ttl-seconds: 21600\") ; \
```
petreeftime, line 11: Should we consider a fallback to IMDSv1 in case v2 isn't available?

petreeftime (approved):

neacsu (approved):

vsock_proxy/service/nitro-enclaves-vsock-proxy.service
andraprs: I was thinking of this, just that the request mentioned to update the "code to only use IMDSv2". We may need to require some exception / share additional info to keep also IMDSv1.

PR #269, opened 2021-07-14
vsock-proxy: Fix remote server's matching against allowlist

aatrand:
Allowlist matching is done by comparing the remote server's IP and port
against the IP and port of all allowlist file entries. The IP-based comparison
is prone to occasional false-negative failures - consider a case where a
hostname may resolve to different IP addresses at different times, due to
load-balancing, for example. The following may occur:

1) The IP of a host is obtained at moment T1, with value IP_1
2) Load balancing occurs and the host is switched with another one that has
   a different IP, IP_2
3) The IP of the host as read from the allowlist is obtained at moment T2 > T1
   and because of 2), it resolves to IP_2
4) IP_1 is compared against IP_2 and the check fails, thus preventing the
   vsock-proxy from launching, even though the hostnames are the same.

This change fixes the problem by always attempting to match hostnames first.
The IP comparison now occurs only if the hostname match fails. This way, load
balancing no longer affects the result since the hostnames don't change.

Signed-off-by: Andrei Trandafir <aatrand@amazon.com>

### DIFF
--- a/vsock_proxy/src/starter.rs
+++ b/vsock_proxy/src/starter.rs
@@ -25,14 +25,14 @@ pub const VSOCK_PROXY_PORT: u32 = 8000;
 /// The most common result type provided by VsockProxy operations.
 pub type VsockProxyResult<T> = Result<T, String>;
 
-/// Checks if the forwarded server is allowed
+/// Checks if the forwarded server is allowed, providing its IP on success.
 pub fn check_allowlist(
-    remote_addr: IpAddr,
+    remote_host: &str,
     remote_port: u16,
     config_file: Option<&str>,
     only_4: bool,
     only_6: bool,
-) -> VsockProxyResult<()> {
+) -> VsockProxyResult<IpAddr> {
     if let Some(config_file) = config_file {
         let mut f = File::open(config_file).map_err(|_| "Could not open the file")?;
 
@@ -45,17 +45,35 @@ pub fn check_allowlist(
             .as_vec()
             .ok_or("No allowlist field")?;
 
+        // Obtain the remote server's IP address.
+        let mut addrs = Proxy::parse_addr(&remote_host, only_4, only_6)
+            .map_err(|err| format!("Could not parse remote address: {}", err))?;
+        let remote_addr = *addrs.get(0).ok_or("No IP address found")?;
+
         for raw_service in services {
+            let addr = raw_service["address"].as_str().ok_or("No address field")?;
             let port = raw_service["port"]
                 .as_i64()
                 .ok_or("No port field or invalid type")?;
             let port = port as u16;
 
-            let addr = raw_service["address"].as_str().ok_or("No address field")?;
-            let addrs = Proxy::parse_addr(addr, only_4, only_6)?;
+            // Start by matching against ports.
+            if port != remote_port {
+                continue;
+            }
+
+            // Attempt to match directly against the allowlisted hostname first.
+            if addr == remote_host {
+                info!("Matched with host name \"{}\" and port \"{}\"", addr, port);
+                return Ok(remote_addr);
+            }
+
+            // If hostname matching failed, attempt to match against IPs.
+            addrs = Proxy::parse_addr(addr, only_4, only_6)?;
             for addr in addrs.into_iter() {
-                if addr == remote_addr && port == remote_port {
-                    return Ok(());
+                if addr == remote_addr {
+                    info!("Matched with host IP \"{}\" and port \"{}\"", addr, port);
+                    return Ok(remote_addr);
                 }
             }
         }
@@ -75,7 +93,7 @@ pub struct Proxy {
 impl Proxy {
     pub fn new(
         local_port: u32,
-        remote_addr: IpAddr,
+        remote_host: &str,
         remote_port: u16,
         num_workers: usize,
         config_file: Option<&str>,
@@ -86,11 +104,15 @@ impl Proxy {
             return Err("Number of workers must not be 0".to_string());
         }
         info!("Checking allowlist configuration");
-        check_allowlist(remote_addr, remote_port, config_file, only_4, only_6)
+        let remote_addr = check_allowlist(remote_host, remote_port, config_file, only_4, only_6)
             .map_err(|err| format!("Error at checking the allowlist: {}", err))?;
-
         let pool = ThreadPool::new(num_workers);
         let sock_type = SockType::Stream;
+
+        info!(
+            "Using IP \"{:?}\" for the given server \"{}\"",
+            remote_addr, remote_host
+        );
         Ok(Proxy {
             local_port,
             remote_addr,
--- a/vsock_proxy/src/main.rs
+++ b/vsock_proxy/src/main.rs
@@ -81,10 +81,6 @@ fn main() -> VsockProxyResult<()> {
         .value_of("remote_addr")
         // This argument is required, so clap ensures it's available
         .unwrap();
-    let remote_addrs = Proxy::parse_addr(&remote_addr, only_4, only_6)
-        .map_err(|err| format!("Could not parse remote address: {}", err))?;
-    let remote_addr = *remote_addrs.get(0).ok_or("No IP address found")?;
-    info!("Using IP {:?} for the given server", remote_addr);
 
     let remote_port = matches
         .value_of("remote_port")
--- a/vsock_proxy/tests/connection_test.rs
+++ b/vsock_proxy/tests/connection_test.rs
@@ -23,7 +23,7 @@ fn vsock_connect(port: u32) -> VsockStream {
 #[test]
 fn test_tcp_connection() {
     // Proxy will translate from port 8000 vsock to localhost port 9000 TCP
-    let addr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+    let addr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)).to_string();
     let mut file = NamedTempFile::new().unwrap();
     file.write_all(
         b"allowlist:\n\
@@ -32,7 +32,7 @@ fn test_tcp_connection() {
     .unwrap();
     let proxy = Proxy::new(
         vsock_proxy::starter::VSOCK_PROXY_PORT,
-        addr,
+        &addr,
         9000,
         2,
         file.path().to_str(),


andraprs (dismissed):

andraprs:
The error in the CI comes from the code formatting via cargo fmt.

```
+ echo '=================== cargo fmt ========================='
=================== cargo fmt =========================
+ make nitro-format
mkdir -p ./build
docker image build -t "nitro_cli:1.0" -f tools/Dockerfile1804.x86_64 tools/
Sending build context to Docker daemon   7.68kB

Step 1/28 : FROM ubuntu:18.04
 ---> ccc6e87d482b
Step 2/28 : SHELL ["/bin/bash", "-c"]
 ---> Using cache
 ---> d898327356a0
Step 3/28 : RUN apt-get update
 ---> Using cache
 ---> a00969154763
Step 4/28 : RUN apt-get install -y curl
 ---> Using cache
 ---> 8fb02eb2d2a2
Step 5/28 : RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ---> Using cache
 ---> 29f0f0604b8e
Step 6/28 : RUN echo "source $HOME/.cargo/env" >> $HOME/.bashrc
 ---> Using cache
 ---> 213bd9c161ff
Step 7/28 : RUN apt-get install -y gcc libssl-dev
 ---> Using cache
 ---> 7289b426e036
Step 8/28 : RUN apt-get install -y pkg-config
 ---> Using cache
 ---> 97554adec33a
Step 9/28 : RUN apt-get install -y musl-tools
 ---> Using cache
 ---> c8ba0a7edbcf
Step 10/28 : RUN apt-get update && apt-get install -y llvm-dev
 ---> Using cache
 ---> 24c6f4ea529a
Step 11/28 : RUN apt-get update && apt-get install -y libclang-dev
 ---> Using cache
 ---> 363cc3c05d79
Step 12/28 : RUN apt-get update && apt-get install -y clang
 ---> Using cache
 ---> 17ea476ba044
Step 13/28 : RUN apt-get install -y wget unzip
 ---> Using cache
 ---> 0ee602631c46
Step 14/28 : ENV OPENSSL_VERSION=OpenSSL_1_1_1d
 ---> Using cache
 ---> c20f3fe0f7c4
Step 15/28 : RUN mkdir /openssl_src
 ---> Using cache
 ---> fbeeaa125ea3
Step 16/28 : RUN wget https://github.com/openssl/openssl/archive/${OPENSSL_VERSION}.zip -P /openssl_src
 ---> Using cache
 ---> acdd147e80c9
Step 17/28 : RUN unzip /openssl_src/${OPENSSL_VERSION}.zip -d /openssl_src
 ---> Using cache
 ---> 3dacd03c1d0a
Step 18/28 : RUN cd /openssl_src/openssl-${OPENSSL_VERSION} && CC=musl-gcc CFLAGS=-fPIC ./Configure --prefix=/musl_openssl --openssldir=/musl_openssl no-shared no-engine no-afalgeng linux-x86_64 -DOPENSSL_NO_SECURE_MEMORY && make && make install
 ---> Using cache
 ---> d88959d1ebfa
Step 19/28 : ENV RUST_VERSION=1.49.0
 ---> Using cache
 ---> d79cf86091eb
Step 20/28 : RUN  source $HOME/.cargo/env &&      rustup toolchain install ${RUST_VERSION}-x86_64-unknown-linux-gnu
 ---> Using cache
 ---> 55fb35a30674
Step 21/28 : RUN  source $HOME/.cargo/env &&     rustup default ${RUST_VERSION}-x86_64-unknown-linux-gnu
 ---> Using cache
 ---> 055cf67c3d5f
Step 22/28 : RUN  source $HOME/.cargo/env &&     rustup target add --toolchain ${RUST_VERSION} x86_64-unknown-linux-musl
 ---> Using cache
 ---> 97a587ca2caa
Step 23/28 : RUN  source $HOME/.cargo/env && 	cargo install cargo-audit --version ^0.15
 ---> Using cache
 ---> 733338243ad5
Step 24/28 : RUN  source $HOME/.cargo/env && 	cargo install cargo-about --version 0.2.2
 ---> Using cache
 ---> 99e0025a11fc
Step 25/28 : RUN apt-get install -y jq
 ---> Using cache
 ---> 1fa2be94539f
Step 26/28 : RUN apt-get update &&     apt-get -y install apt-transport-https         ca-certificates         curl         gnupg2         software-properties-common &&     curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg > /tmp/dkey; apt-key add /tmp/dkey &&     add-apt-repository         "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")         $(lsb_release -cs)         stable" &&     apt-get update &&     apt-get -y install docker-ce
 ---> Using cache
 ---> 2c84b9338c24
Step 27/28 : RUN mkdir -p /var/log/nitro_enclaves
 ---> Using cache
 ---> 167b08df11bf
Step 28/28 : RUN echo "Container build ready to go"
 ---> Using cache
 ---> cd42f491615c
Successfully built cd42f491615c
Successfully tagged nitro_cli:1.0
touch .build-container
docker run \
	-v "$(readlink -f .)":/nitro_src \
	-v "$(readlink -f ./build)":/nitro_build \
	"nitro_cli:1.0" bin/bash -c \
		'source /root/.cargo/env && \
		cargo fmt --manifest-path=/nitro_src/Cargo.toml -q --all -- --check '
Makefile:232: recipe for target 'nitro-format' failed
make: *** [nitro-format] Error 1
```

aatrand:
> The error in the CI comes from the code formatting via cargo fmt.
> 
> ```
> + echo '=================== cargo fmt ========================='
> =================== cargo fmt =========================
> + make nitro-format
> mkdir -p ./build
> docker image build -t "nitro_cli:1.0" -f tools/Dockerfile1804.x86_64 tools/
> Sending build context to Docker daemon   7.68kB
> 
> Step 1/28 : FROM ubuntu:18.04
>  ---> ccc6e87d482b
> Step 2/28 : SHELL ["/bin/bash", "-c"]
>  ---> Using cache
>  ---> d898327356a0
> Step 3/28 : RUN apt-get update
>  ---> Using cache
>  ---> a00969154763
> Step 4/28 : RUN apt-get install -y curl
>  ---> Using cache
>  ---> 8fb02eb2d2a2
> Step 5/28 : RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
>  ---> Using cache
>  ---> 29f0f0604b8e
> Step 6/28 : RUN echo "source $HOME/.cargo/env" >> $HOME/.bashrc
>  ---> Using cache
>  ---> 213bd9c161ff
> Step 7/28 : RUN apt-get install -y gcc libssl-dev
>  ---> Using cache
>  ---> 7289b426e036
> Step 8/28 : RUN apt-get install -y pkg-config
>  ---> Using cache
>  ---> 97554adec33a
> Step 9/28 : RUN apt-get install -y musl-tools
>  ---> Using cache
>  ---> c8ba0a7edbcf
> Step 10/28 : RUN apt-get update && apt-get install -y llvm-dev
>  ---> Using cache
>  ---> 24c6f4ea529a
> Step 11/28 : RUN apt-get update && apt-get install -y libclang-dev
>  ---> Using cache
>  ---> 363cc3c05d79
> Step 12/28 : RUN apt-get update && apt-get install -y clang
>  ---> Using cache
>  ---> 17ea476ba044
> Step 13/28 : RUN apt-get install -y wget unzip
>  ---> Using cache
>  ---> 0ee602631c46
> Step 14/28 : ENV OPENSSL_VERSION=OpenSSL_1_1_1d
>  ---> Using cache
>  ---> c20f3fe0f7c4
> Step 15/28 : RUN mkdir /openssl_src
>  ---> Using cache
>  ---> fbeeaa125ea3
> Step 16/28 : RUN wget https://github.com/openssl/openssl/archive/${OPENSSL_VERSION}.zip -P /openssl_src
>  ---> Using cache
>  ---> acdd147e80c9
> Step 17/28 : RUN unzip /openssl_src/${OPENSSL_VERSION}.zip -d /openssl_src
>  ---> Using cache
>  ---> 3dacd03c1d0a
> Step 18/28 : RUN cd /openssl_src/openssl-${OPENSSL_VERSION} && CC=musl-gcc CFLAGS=-fPIC ./Configure --prefix=/musl_openssl --openssldir=/musl_openssl no-shared no-engine no-afalgeng linux-x86_64 -DOPENSSL_NO_SECURE_MEMORY && make && make install
>  ---> Using cache
>  ---> d88959d1ebfa
> Step 19/28 : ENV RUST_VERSION=1.49.0
>  ---> Using cache
>  ---> d79cf86091eb
> Step 20/28 : RUN  source $HOME/.cargo/env &&      rustup toolchain install ${RUST_VERSION}-x86_64-unknown-linux-gnu
>  ---> Using cache
>  ---> 55fb35a30674
> Step 21/28 : RUN  source $HOME/.cargo/env &&     rustup default ${RUST_VERSION}-x86_64-unknown-linux-gnu
>  ---> Using cache
>  ---> 055cf67c3d5f
> Step 22/28 : RUN  source $HOME/.cargo/env &&     rustup target add --toolchain ${RUST_VERSION} x86_64-unknown-linux-musl
>  ---> Using cache
>  ---> 97a587ca2caa
> Step 23/28 : RUN  source $HOME/.cargo/env && 	cargo install cargo-audit --version ^0.15
>  ---> Using cache
>  ---> 733338243ad5
> Step 24/28 : RUN  source $HOME/.cargo/env && 	cargo install cargo-about --version 0.2.2
>  ---> Using cache
>  ---> 99e0025a11fc
> Step 25/28 : RUN apt-get install -y jq
>  ---> Using cache
>  ---> 1fa2be94539f
> Step 26/28 : RUN apt-get update &&     apt-get -y install apt-transport-https         ca-certificates         curl         gnupg2         software-properties-common &&     curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg > /tmp/dkey; apt-key add /tmp/dkey &&     add-apt-repository         "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")         $(lsb_release -cs)         stable" &&     apt-get update &&     apt-get -y install docker-ce
>  ---> Using cache
>  ---> 2c84b9338c24
> Step 27/28 : RUN mkdir -p /var/log/nitro_enclaves
>  ---> Using cache
>  ---> 167b08df11bf
> Step 28/28 : RUN echo "Container build ready to go"
>  ---> Using cache
>  ---> cd42f491615c
> Successfully built cd42f491615c
> Successfully tagged nitro_cli:1.0
> touch .build-container
> docker run \
> 	-v "$(readlink -f .)":/nitro_src \
> 	-v "$(readlink -f ./build)":/nitro_build \
> 	"nitro_cli:1.0" bin/bash -c \
> 		'source /root/.cargo/env && \
> 		cargo fmt --manifest-path=/nitro_src/Cargo.toml -q --all -- --check '
> Makefile:232: recipe for target 'nitro-format' failed
> make: *** [nitro-format] Error 1
> ```

andraprs (dismissed):

andraprs (dismissed):

petreeftime (changes requested):
Please use allowlist everywhere instead of whitelist (including patch description).

petreeftime (commented):
The commit message talks about AWS KMS, but this affects all host checks. I would not include such specificity in the commit message.

petreeftime (commented):
The new code looks correct, but I am uncertain if I understand the exact issue it's trying to solve.

As far as I understand, the allowlist check would only need to happen once, when vsock-proxy starts. This verifies that the input parameter (which could be an IP or a hostname) and port are part of the allowlist.

Afterwards, on each new connection, vsock-proxy already knows that the address is correct. All it needs to do, in case the allowlist entry is a hostname and not an IP, is ask DNS for an IP value of the host and use the IP it gets to connect to it. As far as I can tell, we re-check the allowlist on each attempt and also try to resolve the IP address to an IP address, which doesn't seem wrong, but also not correct.

EDIT: I think we could also check for the case where the input parameter is an IP address that resolves to a hostname in the allowlist, or the other way around, however, that doesn't seem an immediate concern.

vsock_proxy/src/starter.rs
```diff
@@ -45,17 +45,35 @@ pub fn check_allowlist(
...
+                return Ok(remote_addr);
+            }
+
+            // If hostname matching failed, attempt to match against IPs.
+            addrs = Proxy::parse_addr(addr, only_4, only_6)?;
             for addr in addrs.into_iter() {
-                if addr == remote_addr && port == remote_port {
-                    return Ok(());
+                if addr == remote_addr {
+                    info!("Matched with host IP \"{}\" and port \"{}\"", addr, port);
```
alexggh, line 75: wouldn't this be too verbose ? doesn't it get printed for every packet ?

aatrand: That only happens when the vsock-proxy is launched, and if RUST_LOG=info is set. So it shouldn't get printed for every packet.

vsock_proxy/src/starter.rs
```diff
@@ -45,17 +45,35 @@ pub fn check_allowlist(
             .as_vec()
             .ok_or("No allowlist field")?;
 
+        // Obtain the remote server's IP address.
+        let mut addrs = Proxy::parse_addr(&remote_host, only_4, only_6)
+            .map_err(|err| format!("Could not parse remote address: {}", err))?;
+        let remote_addr = *addrs.get(0).ok_or("No IP address found")?;
```
alexggh, line 51: Why do we check only the first one, shouldn't we check all possible ip-addresses for a given hostname ?

aatrand: This was the behavior before this change and in practice there aren't any reported issues with it, so I didn't see a need to change it.

alexggh: Are we doing this for every packet ? If yes than we might incur a performance penalty, which probably we don't want to pay every packet. 

So what I would do would be:
1. Keep the list of allowed ips cached.
2. If we don't find the IP in the list of allowed ip, refresh the list and check again.
3. Let's also log somewhere when allowlist returns false, so we are able to easily debug issues later on.

aatrand: This is not done for every packet - it's performed only once, when the vsock-proxy starts. If the allowlist matching fails, the vsock-proxy won't start at all.

alexggh (approved):

andraprs (approved):

petreeftime (approved):
The commit is ok in this form and I believe it will help fix the TOCTOU issue described here, however, I would like an additional integration test that provides an allow list and shows that validation happens correctly.

PR #281, opened 2021-08-04
Add `--disconnect-timeout` option to console subcommand

### LINKED ISSUE #229: Feature: Add the capability to disconnect from the console after a set time
There are times (in scripts, etc) that I would really like to run something like this:

`nitro-cli console --enclave-id i-0ee2fsdfdsfdsfdsdf47cce-enc177b7b7sdfsdfdsfdsfs --disconnect 10s`

So, that I can get some output from the console, analyze it and then move on to the next step.

costisin:
*Issue #, if available:*
Resolves #229

*Description of changes:*
Add `--disconnect-timeout` option to console subcommand.
Add cli argument tests and integration tests for the `--disconnect-timeout` option.

### DIFF
--- a/src/common/commands_parser.rs
+++ b/src/common/commands_parser.rs
@@ -163,6 +163,8 @@ impl TerminateEnclavesArgs {
 pub struct ConsoleArgs {
     /// The ID of the enclave whose console is to be shown.
     pub enclave_id: String,
+    /// The time in seconds after the console disconnects from the enclave.
+    pub disconnect_timeout_sec: Option<u64>,
 }
 
 impl ConsoleArgs {
@@ -171,6 +173,8 @@ impl ConsoleArgs {
         Ok(ConsoleArgs {
             enclave_id: parse_enclave_id(args)
                 .map_err(|e| e.add_subaction("Parse enclave ID".to_string()))?,
+            disconnect_timeout_sec: parse_disconnect_timeout(args)
+                .map_err(|e| e.add_subaction("Parse disconnect timeout".to_string()))?,
         })
     }
 }
@@ -315,6 +319,24 @@ fn parse_enclave_id(args: &ArgMatches) -> NitroCliResult<String> {
         )
     })?;
     Ok(enclave_id.to_string())
+}
+
+/// Parse the disconnect timeout from the command-line arguments.
+fn parse_disconnect_timeout(args: &ArgMatches) -> NitroCliResult<Option<u64>> {
+    let disconnect_timeout = match args.value_of("disconnect-timeout") {
+        Some(arg) => Some(arg.parse::<u64>().map_err(|_| {
+            new_nitro_cli_failure!(
+                "`disconnect-timeout` argument can't be parsed as a number",
+                NitroCliErrorEnum::InvalidArgument
+            )
+            .add_info(vec![
+                "disconnect-timeout",
+                args.value_of("disconnect-timeout").unwrap(),
+            ])
+        })?),
+        None => None,
+    };
+    Ok(disconnect_timeout)
 }
 
 /// Parse the list of requested CPU IDs from the command-line arguments.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,15 +216,21 @@ fn artifacts_path() -> NitroCliResult<String> {
 }
 
 /// Wrapper over the console connection function.
-pub fn console_enclaves(enclave_cid: u64) -> NitroCliResult<()> {
+pub fn console_enclaves(
+    enclave_cid: u64,
+    disconnect_timeout_sec: Option<u64>,
+) -> NitroCliResult<()> {
     debug!("console_enclaves");
     println!("Connecting to the console for enclave {}...", enclave_cid);
-    enclave_console(enclave_cid)?;
+    enclave_console(enclave_cid, disconnect_timeout_sec)?;
     Ok(())
 }
 
 /// Connects to the enclave console and prints it continously.
-pub fn enclave_console(enclave_cid: u64) -> NitroCliResult<()> {
+pub fn enclave_console(
+    enclave_cid: u64,
+    disconnect_timeout_sec: Option<u64>,
+) -> NitroCliResult<()> {
     let console = Console::new(
         VMADDR_CID_HYPERVISOR,
         u32::try_from(enclave_cid).map_err(|err| {
@@ -237,7 +243,7 @@ pub fn enclave_console(enclave_cid: u64) -> NitroCliResult<()> {
     .map_err(|e| e.add_subaction("Connect to enclave console".to_string()))?;
     println!("Successfully connected to the console.");
     console
-        .read_to(io::stdout().by_ref())
+        .read_to(io::stdout().by_ref(), disconnect_timeout_sec)
         .map_err(|e| e.add_subaction("Connect to enclave console".to_string()))?;
 
     Ok(())
@@ -462,6 +468,13 @@ macro_rules! create_app {
                             .takes_value(true)
                             .help("Enclave ID, used to uniquely identify an enclave")
                             .required(true),
+                    )
+                    .arg(
+                        Arg::with_name("disconnect-timeout")
+                            .long("disconnect-timeout")
+                            .takes_value(true)
+                            .help("The time in seconds after the console disconnects from the enclave")
+                            .required(false)
                     ),
             )
             .subcommand(
--- a/src/main.rs
+++ b/src/main.rs
@@ -222,7 +222,8 @@ fn main() {
                 })
                 .ok_or_exit_with_errno(None);
             }
-            console_enclaves(enclave_cid)
+
+            console_enclaves(enclave_cid, console_args.disconnect_timeout_sec)
                 .map_err(|e| {
                     e.add_subaction("Failed to connect to enclave console".to_string())
                         .set_action(ENCLAVE_CONSOLE_STR.to_string())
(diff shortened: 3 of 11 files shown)


src/common/commands_parser.rs
```diff
@@ -163,6 +163,8 @@ impl TerminateEnclavesArgs {
 pub struct ConsoleArgs {
     /// The ID of the enclave whose console is to be shown.
     pub enclave_id: String,
+    /// The time in seconds after the console disconnects from the enclave.
+    pub disconnect_timeout: Option<u64>,
```
alexggh, line 167: then let's call it disconnect_timeout_ms.

src/lib.rs
```diff
@@ -236,6 +239,15 @@ pub fn enclave_console(enclave_cid: u64) -> NitroCliResult<()> {
     )
     .map_err(|e| e.add_subaction("Connect to enclave console".to_string()))?;
     println!("Successfully connected to the console.");
+
+    // Start the timeout only if the option was provided
+    if disconnect_timeout != None {
+        thread::spawn(move || {
+            thread::sleep(Duration::from_secs(disconnect_timeout.unwrap()));
+            process::exit(0);
```
alexggh, line 247: Neah, this not the right way to terminate the console, we should have a read_to_with_timeout that exits when the timeout expires.

src/lib.rs
```diff
@@ -462,6 +474,13 @@ macro_rules! create_app {
                             .takes_value(true)
                             .help("Enclave ID, used to uniquely identify an enclave")
                             .required(true),
+                    )
+                    .arg(
+                        Arg::with_name("disconnect")
```
alexggh, line 479: disconnect-after, is much more expressive.

src/main.rs
```diff
@@ -222,7 +222,9 @@ fn main() {
                 })
                 .ok_or_exit_with_errno(None);
             }
-            console_enclaves(enclave_cid)
+            let disconnect_timeout = console_args.disconnect_timeout;
```
alexggh, line 225: you don't need this let.

tests/tests.rs
```diff
@@ -473,7 +473,7 @@ mod tests {
         let replies: Vec<EnclaveDescribeInfo> = vec![info];
         let _reply = &replies[0];
 
-        assert_eq!(enclave_console(enclave_cid).is_err(), true);
+        assert_eq!(enclave_console(enclave_cid, 0).is_err(), true);
```
alexggh: shouldn't it be None ? I fail to see how this compiles.

tests/tests.rs
```diff
@@ -473,7 +473,7 @@ mod tests {
         let replies: Vec<EnclaveDescribeInfo> = vec![info];
         let _reply = &replies[0];
 
-        assert_eq!(enclave_console(enclave_cid, 0).is_err(), true);
+        assert_eq!(enclave_console(enclave_cid, None).is_err(), true);
```
alexggh, line 476: should be solved in the initial commit that introduces the problem.

alexggh (changes requested):

src/common/commands_parser.rs
costisin: why ms? the time is measured in seconds

alexggh: seconds :)

alexggh: no, this was not resolved disconnect_timeout_sec.

src/lib.rs
```diff
@@ -236,8 +236,9 @@ pub fn enclave_console(enclave_cid: u64) -> NitroCliResult<()> {
     )
     .map_err(|e| e.add_subaction("Connect to enclave console".to_string()))?;
     println!("Successfully connected to the console.");
+
```
alexggh, line 239: nit: unrelated white space

src/lib.rs
alexggh: no, it was not resolved.

src/utils.rs
```diff
@@ -141,8 +142,54 @@ impl Console {
...
+                    NitroCliErrorEnum::EpollError
+                )
+            },
+        )?;
+
+        let mut events = [EpollEvent::empty(); 1];
+        // Disconnect timeout in microseconds
+        let mut epoll_timeout_us = match disconnect_timeout {
+            Some(time) => time as i128 * 1000 * 1000,
+            None => -1 * 1000,
```
alexggh, line 171: so, you want -1 here.
Specifying a timeout of  -1  causes
       epoll_wait() to block indefinitely, while specifying a timeout equal to
       zero cause epoll_wait() to return immediately, even if  no  events  are
       available.

src/utils.rs
```diff
@@ -166,8 +213,27 @@ impl Console {
                     )
                 })?;
             }
+
+            // Account for the read/write/epoll_wait elapsed time
+            epoll_timeout_us -= start_epoll_time.elapsed().as_micros() as i128;
```
alexggh, line 218: this has unpredictable behavior if timeout is not sent.

src/utils.rs
```diff
@@ -166,8 +213,27 @@ impl Console {
...
+        // Clear console read fd from epoll
+        epoll::epoll_ctl(epoll_fd, EpollOp::EpollCtlDel, self.fd, None).map_err(|e| {
+            new_nitro_cli_failure!(
+                &format!("Failed to remove fd from epoll: {:?}", e),
+                NitroCliErrorEnum::EpollError
+            )
+        })?;
+
+        // Close epoll fd
+        unistd::close(epoll_fd).map_err(|e| {
```
alexggh, line 230: I don't expect this is to be needed, why do you think is needed.
Normally the ownership of the underlying filedescriptor it is owned by the EpollFd struct and when that goes out of scope, it will close the fd as well.

src/utils.rs
costisin: this variable is stored in microseconds, but later in the code #L180, this is conveted to milliseconds (* 1000), so after conversion, the -1000 will become -1

alexggh (dismissed):

src/utils.rs
```diff
@@ -141,8 +142,54 @@ impl Console {
     }
 
     /// Read a chunk of raw data from the console and output it.
-    pub fn read_to(&self, output: &mut dyn Write) -> NitroCliResult<()> {
+    pub fn read_to(
+        &self,
+        output: &mut dyn Write,
+        disconnect_timeout_sec: Option<u64>,
+    ) -> NitroCliResult<()> {
+        let epoll_fd = epoll::epoll_create().map_err(|e| {
```
aatrand, line 150: The epoll logic should run only if ```disconnect_timeout_sec``` is not None. Otherwise we affect the performance of the ```read_to()``` function when it's called without a disconnect interval.

src/utils.rs
```diff
@@ -166,8 +213,29 @@ impl Console {
...
             }
+
+            // Account for the read/write/epoll_wait elapsed time
+            if disconnect_timeout_sec.is_some() {
+                epoll_timeout_us -= start_epoll_time.elapsed().as_micros() as i128;
+            }
         }
 
+        // Clear console read fd from epoll
+        epoll::epoll_ctl(epoll_fd, EpollOp::EpollCtlDel, self.fd, None).map_err(|e| {
```
aatrand, line 224: This function will never be called if the ```read()``` or ```write()``` functions inside the loop fail. While this isn't critical at the moment, since the failure will immediately lead to the termination of the nitro-cli, we should ensure that we remove the epoll descriptor no matter what.

src/utils.rs
```diff
@@ -141,14 +142,90 @@ impl Console {
     }
 
     /// Read a chunk of raw data from the console and output it.
-    pub fn read_to(&self, output: &mut dyn Write) -> NitroCliResult<()> {
+    pub fn read_to(
+        &self,
+        output: &mut dyn Write,
+        disconnect_timeout_sec: Option<u64>,
+    ) -> NitroCliResult<()> {
+        fn epoll_cleanup(epoll_fd: i32, fd: i32) -> NitroCliResult<()> {
```
alexggh, line 150: no, please do not do this, it makes the code way harder to read. make it a  separate function if you really need to.
the right type for the parameters is RawFd.

src/utils.rs
```diff
@@ -157,15 +234,28 @@ impl Console {
 
             if size > 0 {
                 output.write(&buffer[..size]).map_err(|e| {
-                    new_nitro_cli_failure!(
-                        &format!(
-                            "Failed to write data from the console to the given stream: {:?}",
-                            e
+                    let cleanup_result = epoll_cleanup(epoll_fd, self.fd);
+                    match cleanup_result {
```
alexggh, line 238: why do you need this match ?

src/utils.rs
alexggh: Could we please just use this: https://docs.rs/vmm-sys-util/0.8.0/vmm_sys_util/eventfd/struct.EventFd.html so that we  don't really care about cleanup.

src/utils.rs
costisin: because the result of the `epoll_cleanup` function must be consumed

src/utils.rs
costisin: You are right with the parameter types. 
`epoll_cleanup` is only relevant in this context, why make a new function in this utils module, if it will only be used in this context?
We can't use EventFd because it's not only about closing the epoll socket, but more about making the epoll_ctl call with EpollCtlDel flag.

src/utils.rs
alexggh: why does it need to be consumed ?

costisin: because in Rust there is a rule: If your function returns a Result<T, E> you must consume it.

alexggh (approved):

aatrand (approved):

PR #287, opened 2021-08-18
eif_defs: Add new metadata EIF section type

raulmldv:
Added EifSectionMetadata to the EifSectionType enum. The
purpose is to use it in eVMM to update the EIF version.
V4 will contain the metadata section added the with the
`build-enclaves` command.

Also added Metadata struct for (de)serializing the EIF section
in eVMM. Currently treating EifSectionMetadata as an invalid
section in EifReader until we merge the rest of the metadata
implementation.  

### DIFF
--- a/eif_defs/src/lib.rs
+++ b/eif_defs/src/lib.rs
@@ -14,7 +14,8 @@ use byteorder::{BigEndian, ByteOrder};
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 use serde::{Deserialize, Serialize};
-use std::mem::size_of;
+use serde_json::Value;
+use std::{collections::BTreeMap, mem::size_of};
 
 pub const EIF_MAGIC: [u8; 4] = [46, 101, 105, 102]; // .eif in ascii
 pub const MAX_NUM_SECTIONS: usize = 32;
@@ -168,6 +169,7 @@ pub enum EifSectionType {
     EifSectionCmdline,
     EifSectionRamdisk,
     EifSectionSignature,
+    EifSectionMetadata,
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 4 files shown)


axlprv (changes requested):
We should also define here the structures used to (de)serialize the metadata section.

eif_defs/src/lib.rs
```diff
@@ -250,6 +252,42 @@ impl PcrInfo {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "CustomMetadata")]
+    /// Metadata provided by the user
+    pub custom_meta: Option<Value>,
+}
+
+impl Metadata {
+    pub fn new(
+        img_name: String,
+        img_version:String,
```
costisin, line 276:
```suggestion
        img_version: String,
```

eif_defs/src/lib.rs
```diff
@@ -250,6 +252,42 @@ impl PcrInfo {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Metadata {
+    #[serde(rename = "ImageName")]
+    pub img_name: String,
+    #[serde(rename = "ImageVersion")]
+    pub img_version:String,
```
costisin, line 260: you could also add comments for this two fields, just for consistency

costisin (approved):

eif_defs/src/lib.rs
```diff
@@ -341,7 +381,7 @@ mod tests {
     #[test]
     fn test_eif_section_header_from_be_bytes_invalid_section_type() {
         let mut bytes = [0u8; 12];
-        bytes[1] = 5;
+        bytes[1] = 6;
```
lexnv, line 384: We could add a small comment explaining the logic, as to why this is invalid.

lexnv (dismissed):

eif_defs/src/lib.rs
raulmldv: Sure, added.

andraprs (approved):

axlprv (approved):

PR #294, opened 2021-08-31
Named enclaves

raulmldv:
### User friendly names for enclaves

Introducing a new option for the `run-enclaves` command allowing users to set custom names for running enclaves instead of using only the ID. The name is set following the option `--enclave-name` and will be uniquely assigned to each of the running enclaves. In case of name conflicts, an index will be concatenated to the initial name, incrementing until the name becomes unique. This action is optional and if no name is provided, the EIF name will take its place.
The name will be part of the output of the following commands: `run-enclave`, `describe-enclaves` and `terminate-enclave`.

Commands that use the ID for targeting a specific enclave, now have the option to address the enclave by its name. So the commands `console` and `terminate-enclave` will have the option to use `--enclave-name` instead of `--enclave-id`.

Unique name assigning is done by the CLI process by broadcasting a request for all the running enclaves' names. The new name is checked agains all the others and saved in the EnclaveManager or changed if needed (with the index added) . Name addressing for the other commands is done by sending, again, to all te running enclaves, the name provided by the user. The matching enclave will send its ID and the command will be executed as before, using the ID.

### Command outputs

```
$ ./nitro-cli run-enclave --cpu-count 2 --memory 256 --enclave-cid 16 --eif-path hello.eif --enclave-name testName --debug-mode
Start allocating memory...
Started enclave with enclave-cid: 16, memory: 256 MiB, cpu-ids: [1, 9]
{
  "EnclaveName": "testName",
  "EnclaveID": "i-0584873aac8fea23b-enc17b9c558602bf9f",
  "ProcessID": 652683,
  "EnclaveCID": 16,
  "NumberOfCPUs": 2,
  "CPUIDs": [
    1,
    9
  ],
  "MemoryMiB": 256
}

$ ./nitro-cli describe-enclaves
[
  {
    "EnclaveName": "testName",
    "EnclaveID": "i-0584873aac8fea23b-enc17b9c558602bf9f",
    "ProcessID": 652683,
    "EnclaveCID": 16,
    "NumberOfCPUs": 2,
    "CPUIDs": [
      1,
      9
    ],
    "MemoryMiB": 256,
    "State": "RUNNING",
    "Flags": "DEBUG_MODE",
    "Measurements": {
      "HashAlgorithm": "Sha384 { ... }",
      "PCR0": "3eaa317a778ee207fab894f6b42c5545ddef8bd905da8cc2f22cc59cc83fe0c980a411e5f5445f110f26d5c0d6f39ee4",
      "PCR1": "aca6e62ffbf5f7deccac452d7f8cee1b94048faf62afc16c8ab68c9fed8c38010c73a669f9a36e596032f0b973d21895",
      "PCR2": "ea436e643c31f2891161803e3f22084a12d38f807ef733379cadf789ec85d51e3bb787f48f3cad0c291fbc82fe7ca27d"
    }
  }
]

$ ./nitro-cli console --enclave-name testName
Connecting to the console for enclave 16...
Successfully connected to the console.
[    0.000000] Linux version 4.14.177-104.253.amzn2.x86_64 (mockbuild@ip-10-0-1-32) (gcc version 7.3.1 20180712 (Red Hat 7.3.1-6) (GCC)) #1 SMP Fri May 1 02:01:13 UTC 2020
..........

```
In the scenario that two enclaves could run simultaneously, a second run command with the exact name will result:

```
$ ./nitro-cli run-enclave --cpu-count 2 --memory 256 --enclave-cid 16 --eif-path hello.eif --enclave-name testName --debug-mode
Start allocating memory...
Started enclave with enclave-cid: 16, memory: 256 MiB, cpu-ids: [1, 9]
{
  "EnclaveName": "testName_1",
  "EnclaveID": "i-0584873aac8fea23b-enc17b9c55other_id",
  "ProcessID": 652683,
  "EnclaveCID": 16,
  "NumberOfCPUs": 2,
  "CPUIDs": [
    1,
    9
  ],
  "MemoryMiB": 256
}
```

```
$ ./nitro-cli terminate-enclave --enclave-name testName
Successfully terminated enclave i-0584873aac8fea23b-enc17b9c558602bf9f.
{
  "EnclaveName": "testName",
  "EnclaveID": "i-0584873aac8fea23b-enc17b9c558602bf9f",
  "Terminated": true
}
```

### DIFF
--- a/src/common/commands_parser.rs
+++ b/src/common/commands_parser.rs
@@ -11,6 +11,7 @@ use serde::{Deserialize, Serialize};
 use std::fs::File;
 
 use crate::common::{NitroCliErrorEnum, NitroCliFailure, NitroCliResult, VMADDR_CID_PARENT};
+use crate::get_id_by_name;
 use crate::new_nitro_cli_failure;
 
 /// The arguments used by the `run-enclave` command.
@@ -155,10 +156,19 @@ pub struct TerminateEnclavesArgs {
 impl TerminateEnclavesArgs {
     /// Construct a new `TerminateEnclavesArgs` instance from the given command-line arguments.
     pub fn new_with(args: &ArgMatches) -> NitroCliResult<Self> {
-        Ok(TerminateEnclavesArgs {
-            enclave_id: parse_enclave_id(args)
-                .map_err(|e| e.add_subaction("Parse enclave ID".to_string()))?,
-        })
+        // If a name is given, find the corresponding EnclaveID
+        match parse_enclave_name(args)
+            .map_err(|e| e.add_subaction("Parse Enclave Name".to_string()))?
+        {
+            Some(name) => Ok(TerminateEnclavesArgs {
+                enclave_id: get_id_by_name(name)
+                    .map_err(|e| e.add_subaction("Get ID by Name".to_string()))?,
+            }),
+            None => Ok(TerminateEnclavesArgs {
+                enclave_id: parse_enclave_id(args)
+                    .map_err(|e| e.add_subaction("Parse enclave ID".to_string()))?,
+            }),
+        }
     }
 }
 
@@ -174,9 +184,19 @@ pub struct ConsoleArgs {
 impl ConsoleArgs {
     /// Construct a new `ConsoleArgs` instance from the given command-line arguments.
     pub fn new_with(args: &ArgMatches) -> NitroCliResult<Self> {
-        Ok(ConsoleArgs {
-            enclave_id: parse_enclave_id(args)
+        // If a name is given, find the corresponding EnclaveID
+        let enclave_id = match parse_enclave_name(args)
+            .map_err(|e| e.add_subaction("Parse Enclave Name".to_string()))?
+        {
+            Some(name) => {
+                get_id_by_name(name).map_err(|e| e.add_subaction("Get ID by Name".to_string()))?
+            }
+            None => parse_enclave_id(args)
                 .map_err(|e| e.add_subaction("Parse enclave ID".to_string()))?,
+        };
+
+        Ok(ConsoleArgs {
+            enclave_id,
             disconnect_timeout_sec: parse_disconnect_timeout(args)
                 .map_err(|e| e.add_subaction("Parse disconnect timeout".to_string()))?,
         })
--- a/src/common/commands_parser.rs
+++ b/src/common/commands_parser.rs
@@ -28,6 +28,8 @@ pub struct RunEnclavesArgs {
     pub debug_mode: Option<bool>,
     /// The number of CPUs that the enclave will receive.
     pub cpu_count: Option<u32>,
+    /// Enclave name set by the user.
+    pub enclave_name: Option<String>,
 }
 
 impl RunEnclavesArgs {
@@ -75,6 +77,8 @@ impl RunEnclavesArgs {
                 cpu_ids: parse_cpu_ids(args)
                     .map_err(|err| err.add_subaction("Parse CPU IDs".to_string()))?,
                 debug_mode: debug_mode(args),
+                enclave_name: parse_enclave_name(args)
+                    .map_err(|err| err.add_subaction("Parse enclave name".to_string()))?,
             })
         }
     }
@@ -390,6 +394,11 @@ fn debug_mode(args: &ArgMatches) -> Option<bool> {
     } else {
         None
     }
+}
+
+/// Parse the enclave name from the command-line arguments.
+fn parse_enclave_name(args: &ArgMatches) -> NitroCliResult<Option<String>> {
+    Ok(args.value_of("enclave-name").map(|e| e.to_string()))
 }
 
 fn parse_signing_certificate(args: &ArgMatches) -> Option<String> {
--- a/src/common/json_output.rs
+++ b/src/common/json_output.rs
@@ -65,19 +65,28 @@ impl EnclaveDescribeInfo {
 /// like measurements, not found in older NitroCLI versions
 #[derive(Clone, Serialize, Deserialize)]
 pub struct DescribeOutput {
+    /// Enclave name assigned by the user
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "EnclaveName")]
+    pub enclave_name: Option<String>,
     #[serde(flatten)]
     /// General describe info found in all versions of NitroCLI
     describe_info: EnclaveDescribeInfo,
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(flatten)]
     /// Build measurements containing PCRs
-    build_info: Option<EnclaveBuildInfo>,
+    pub build_info: Option<EnclaveBuildInfo>,
 }
 
 impl DescribeOutput {
     /// Creates new describe output from available
-    pub fn new(describe_info: EnclaveDescribeInfo, build_info: Option<EnclaveBuildInfo>) -> Self {
+    pub fn new(
+        enclave_name: Option<String>,
+        describe_info: EnclaveDescribeInfo,
+        build_info: Option<EnclaveBuildInfo>,
+    ) -> Self {
         DescribeOutput {
+            enclave_name,
             describe_info,
             build_info,
         }
@@ -87,6 +96,9 @@ impl DescribeOutput {
 /// The information to be provided for a `run-enclave` request.
 #[derive(Clone, Serialize, Deserialize)]
 pub struct EnclaveRunInfo {
+    #[serde(rename = "EnclaveName")]
+    /// The name of the enclave.
+    pub enclave_name: String,
     #[serde(rename = "EnclaveID")]
     /// The full ID of the enclave.
     pub enclave_id: String,
@@ -110,6 +122,7 @@ pub struct EnclaveRunInfo {
 impl EnclaveRunInfo {
     /// Create a new `EnclaveRunInfo` instance from the given enclave information.
     pub fn new(
+        enclave_name: String,
         enclave_id: String,
         enclave_cid: u64,
         cpu_count: usize,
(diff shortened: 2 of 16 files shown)


src/common/commands_parser.rs
```diff
@@ -392,6 +396,16 @@ fn debug_mode(args: &ArgMatches) -> Option<bool> {
...
 }
 
+/// Parse the enclave name from the command-line arguments.
+fn parse_enclave_name(args: &ArgMatches) -> NitroCliResult<Option<String>> {
+    let enclave_name = if let Some(enclave_name) = args.value_of("enclave-name") {
+        Some(enclave_name.to_string())
+    } else {
+        None
+    };
+    Ok(enclave_name)
```
costisin, line 427:
```suggestion
    Ok(args.value_of("eif-path").map(|e| e.to_string()))
```

src/common/json_output.rs
```diff
@@ -76,8 +80,13 @@ pub struct DescribeOutput {
 
 impl DescribeOutput {
     /// Creates new describe output from available
-    pub fn new(describe_info: EnclaveDescribeInfo, build_info: Option<EnclaveBuildInfo>) -> Self {
+    pub fn new(
+        enclave_name: Option<String>,
```
costisin, line 84: Why don't you include enclave_name in the EnclaveDescribeInfo struct?

src/enclave_proc/commands.rs
```diff
@@ -48,12 +48,24 @@ pub fn run_enclaves(
         .map_err(|e| e.add_subaction("Failed to construct CPU information".to_string()))?
         .get_cpu_config(args)
         .map_err(|e| e.add_subaction("Failed to get CPU configuration".to_string()))?;
+
+    let enclave_name = match args.enclave_name.clone() {
```
costisin, line 52: Is cloning necessary?

src/lib.rs
```diff
@@ -424,6 +424,41 @@ pub fn get_all_enclave_names() -> NitroCliResult<Vec<String>> {
...
+            NitroCliErrorEnum::EnclaveNamingError
+        ));
+    } else if objects.len() > 1 {
+        return Err(new_nitro_cli_failure!(
+            "Conflicting enclave names have been found.".to_string(),
+            NitroCliErrorEnum::EnclaveNamingError
+        ));
+    }
+
+    Ok(objects[0].clone())
```
costisin, line 459: Why not just moving?

src/lib.rs
```diff
@@ -591,14 +637,23 @@ macro_rules! create_app {
                             .long("enclave-id")
                             .takes_value(true)
                             .help("Enclave ID, used to uniquely identify an enclave")
-                            .required(true),
+                            .required(true)
```
costisin, line 640: You are looking for `required_unless("enclave-name")` which requires an argument unless other argument `enclave-name` is found

src/lib.rs
```diff
@@ -424,6 +424,41 @@ pub fn get_all_enclave_names() -> NitroCliResult<Vec<String>> {
...
+        return Err(new_nitro_cli_failure!(
+            "No enclave matched the given name.".to_string(),
+            NitroCliErrorEnum::EnclaveNamingError
+        ));
+    } else if objects.len() > 1 {
+        return Err(new_nitro_cli_failure!(
+            "Conflicting enclave names have been found.".to_string(),
+            NitroCliErrorEnum::EnclaveNamingError
+        ));
+    }
```
costisin, line 457: Maybe you could rewrite like this so there is no duplicating code
```suggestion
    if (objects.len() != 1) {
        return Err(new_nitro_cli_failure!(
            match objects.len() {
                0 => "No enclave matched the given name.".to_string(),
                2.. => "Conflicting enclave names have been found.".to_string(),
            },
            NitroCliErrorEnum::EnclaveNamingError
        ));
    }
```

src/common/commands_parser.rs
raulmldv: will change with "enclave-name" instead of "eif-path"

src/common/json_output.rs
raulmldv: I avoided changing the returned type by the enclave for keeping the backwards compatibility. I will look for a way of refactoring to avoid confusion.

src/enclave_proc/commands.rs
raulmldv: could use to_string later and borrowing here

src/lib.rs
raulmldv: Cannot move out of vector index, could use remove() instead.

src/common/commands_parser.rs
costisin: yeah, sorry, i meant "enclave-name"

src/lib.rs
```diff
@@ -155,6 +157,40 @@ pub fn build_from_docker(
...
+            .set_action("Get Enclaves Name".to_string())
+    })?;
+
+    let mut idx = 0;
+    let mut result_name = enclave_name.clone();
+    names.sort();
+
+    // If duplicates are found, add index to name eg. testName -> testName_1 -> testName_2 ..
+    for name in names {
+        if result_name == name {
```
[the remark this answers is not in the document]
raulmldv: Once an enclave is assigned a name with the index, it will be added to this array. I realised that it is not guaranteed that I will find all the existing names with indexes in consecutive order in the sorted array (eg. list is: name, nameother, name_1 ...) so I should check this in another way.

src/common/json_output.rs
```diff
@@ -35,20 +39,27 @@ pub struct EnclaveDescribeInfo {
     #[serde(rename = "Flags")]
     /// The bit-mask which provides the enclave's launch flags.
     pub flags: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(flatten)]
+    /// Build measurements containing PCRs
+    pub build_info: Option<EnclaveBuildInfo>,
```
petreeftime, line 45: This seems to be part of another patch.

src/common/json_output.rs
```diff
@@ -35,20 +39,27 @@ pub struct EnclaveDescribeInfo {
...
     pub fn new(
+        enclave_name: Option<String>,
         enclave_id: String,
         enclave_cid: u64,
         cpu_count: u64,
         cpu_ids: Vec<u32>,
         memory_mib: u64,
         state: String,
         flags: String,
+        build_info: Option<EnclaveBuildInfo>,
```
petreeftime, line 59: This seems to be part of another patch.

src/enclave_proc/commands.rs
```diff
@@ -48,12 +48,24 @@ pub fn run_enclaves(
...
+
+    let enclave_name = match &args.enclave_name {
+        Some(enclave_name) => enclave_name,
+        None => {
+            return Err(new_nitro_cli_failure!(
+                "Failed to set enclave name in the manager".to_string(),
+                NitroCliErrorEnum::FileOperationFailure
+            ))
+        }
+    };
```
petreeftime, line 60: Perhaps I'm missing something, but does that mean that enclave_name is not an optional argument? I don't think it should be a required argument. Perhaps setting a useful default is better (eg, the enclave ID or https://github.com/fnichol/names).

petreeftime (commented):
How backwards compatible is this change? If I start an enclave on the current published version and then upgrade the tooling to this, will it still work, or will something crash?

src/common/json_output.rs
raulmldv: > How backwards compatible is this change? If I start an enclave on the current published version and then upgrade the tooling to this, will it still work, or will something crash?

I tested backwards compatibility and developed the feature in this way. The output of the current published version will be missing the additional field of the describe output which will be interpreted by the new CLI process as being `None`, so nothing crashes.

src/enclave_proc/commands.rs
raulmldv: There is a default value to which the name is set in case the argument is missing, the name of the EIF (test.eif -> test). Still, this field is kept as an `Option` because at the moment of parsing the arguments it could be missing.

src/common/json_output.rs
raulmldv: This is added so the output of the `describe-enclaves` includes the new information. Should this be part of another commit or another PR?

src/lib.rs
```diff
@@ -155,6 +157,32 @@ pub fn build_from_docker(
...
+/// Creates new enclave name
+///
+/// Requests the names of all running instances and checks the
+/// occurrence of the chosen name for the new enclave.
+pub fn new_enclave_name(run_args: RunEnclavesArgs, names: Vec<String>) -> NitroCliResult<String> {
+    let enclave_name = match run_args.enclave_name {
+        Some(enclave_name) => enclave_name,
+        None => {
+            // Get name of EIF file from path eg. path/to/eif/hello.eif -> hello
+            let path_split: Vec<&str> = run_args.eif_path.split(&['/', '.'][..]).collect();
```
axlprv, line 169: We should not assume that all the EIF paths have the `.eif` extension as we do not enforce this rule anywhere.

src/common/json_output.rs
petreeftime: Whichever you think makes more sense. I prefer smaller/more PRs.

src/lib.rs
raulmldv: Solved to treat the case of no extension present.

src/common/json_output.rs
raulmldv: I'll keep that in mind for future PRs, for now I added a new commit because it makes more sense with the suggestion of @costisin in this PR.

axlprv (approved):

src/lib.rs
```diff
@@ -419,6 +419,39 @@ pub fn get_all_enclave_names() -> NitroCliResult<Vec<String>> {
     Ok(objects)
 }
 
+/// Sends the name to all the running enclaves and expects a response
+/// with the ID of the one that uniquely matched
```
petreeftime, line 423: What guarantees that names are unique? I haven't seen any check in run-enclave that ensures this.

petreeftime: I've searched a bit more and finally found it.

petreeftime (approved):

PR #298, opened 2021-09-08
 linuxkit: Update linuxkit to v0.8+

lexnv:
The linuxkit blob is updated from the upstream branch
with commit 46d4edc [1].

The blob is slight modified to support prefixing the
filesystem of the docker image. This is necessary for
generating the initrd image.

To avoid size limitations present in the linux kernel,
the gzip wrapper is removed.

### X86_64
- `file ./blobs/x86_64/linuxkit`

```
[Before] dynamically linked, stripped
[After]  statically linked, stripped
```

- file size
```
[Before] 57M
[After]  48M
```

### Aarch64

- `file ./blobs/aarch64/linuxkit`

```
[Before] dynamically linked, stripped
[After]  statically linked, stripped
```

- `file size: 55M`

```
[Before] 55M
[After]  45M
```

[1] https://github.com/linuxkit/linuxkit/commit/46d4edc967b5a9de705840d30c83b83fedabb492

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -32,6 +32,13 @@ mod tests {
     // Remote Docker image
     const SAMPLE_DOCKER: &str =
         "667861386598.dkr.ecr.us-east-1.amazonaws.com/enclaves-samples:vsock-sample";
+    const SAMPLE_DOCKER_PCR0: &str =
+        "93a8a6a775cd1e1ab8b6121d1b0ce08e99e0976dabfa40663fa8ea9633421305de18c8f95aa2b82d3feb918fe912c838";
+    const SAMPLE_DOCKER_PCR1: &str =
+        "c35e620586e91ed40ca5ce360eedf77ba673719135951e293121cb3931220b00f87b5a15e94e25c01fecd08fc9139342";
+    const SAMPLE_DOCKER_PCR2: &str =
+        "52528ebeccf82b21cea3f3a9d055f1bb3d18254d77dcda2bbd7f39cecd96b7eea842913800cc1b0bc261b7ad1b83be90";
+
     // Local Docker image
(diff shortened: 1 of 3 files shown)


petreeftime:
Instead of pushing this to the repo, can we put it somewhere in S3 instead?

tests/tests.rs
```diff
@@ -87,15 +87,15 @@ mod tests {
         .1;
         assert_eq!(
             measurements.get("PCR0").unwrap(),
-            "93a8a6a775cd1e1ab8b6121d1b0ce08e99e0976dabfa40663fa8ea9633421305de18c8f95aa2b82d3feb918fe912c838"
+            "a5a11ad2b3a60117c14b9fe1278b4b902ac0407e76a7b2d7a1f9e60dbacbd8a621acfe080c834c291fd13b348af0c5b1"
```
petreeftime, line 90: Can you explain why the PCRs changed here? I don't think rebuilding the same container image should ever change the PCRs.

petreeftime (changes requested):
The build tool should not influence the PCR values. If the internal layout changed, we should fix it, such that the update is transparent to users.

tests/tests.rs
lexnv: I'll have a look at the changes from upstream to determine a viable path forward for updating the linuxkit. Indeed, the PCRs should not change

shtaked: There are 2 changes I see in the `.img` file produced by newer version of linuxkit:
* the output is `gzip`ped with the same set of arguments we used before
* the content of the image changed - there are some duplicate entries for `rootfs/etc/hosts` and `rootfs/etc/resolv.conf`

shtaked:
Updating the Linuxkit version won't fix the "Out of memory" issues and doesn't worth the effort. Closing the PR.

PR #301, opened 2021-09-10
EIF metadata section and commands

raulmldv:
### EIF v4 - metadata

In completion to the defining of the metadata section in: https://github.com/aws/aws-nitro-enclaves-cli/commit/efd33456d31ef11492974dcd2364981fe102c27a

Changes to the `build-enclave` command have been made, to add a name and version to an EIF and optionally custom metadata in form of a JSON file provided by the user. Default values are set to the image in case no name or version is provided (the docker image uri and tag). This new information forms the new section of the EIF, added at the end of it. The command options used to add metadata are: `--name`, `--version` and '--metadata'.

Metadata consists of the following categories:
- image name and version
- auto-generated metadata -> added automatically at build (build time, build tool, OS)
- docker information -> extracted from the docker image used to build the EIF
- custom metadata -> added by the user with a JSON file argument

This information can be displayed when executing `describe-eif`  or `describe-enclaves` with the option `--metadata` on an image with version >=4. The change is backwards compatible and will not display anything besides the usual output.

With the adding of metadata, some new fields have been added to `describe-eif`'s output: `IsSigned` and `ChackCRC`.
These are the result of both checking the CRC by recalculating and verifying the signature present in the signature section of the EIF.

Note: metadata section is added to the checksum.


eif_utils/src/lib.rs
```diff
@@ -26,11 +28,17 @@ use std::collections::BTreeMap;
...
 use std::mem::size_of;
 use std::path::Path;
+use std::time::SystemTime;
 
 pub const DEFAULT_SECTIONS_COUNT: u16 = 2;
+pub const NITROCLI_SPECS_PATH: &str = "../../SPECS/aws-nitro-enclaves-cli.spec";
+#[cfg(target_arch = "x86_64")]
+pub const KERNEL_CONFIG_PATH: &str = "../../blobs/x86_64/bzImage.config";
+#[cfg(target_arch = "aarch64")]
+pub const KERNEL_CONFIG_PATH: &str = "../../blobs/aarch64/bzImage.config";
```
popegeo: for aarch64 the config file is not called bzImage.config anymore

eif_utils/src/lib.rs
```diff
@@ -711,8 +712,8 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
         meta.insert("BuildTool".to_string(), tool_name);
         meta.insert("BuildToolVersion".to_string(), tool_version);
 
-        let config_file =
-            File::open(KERNEL_CONFIG_PATH).expect("Failed to open kernel image config file.");
+        let config_file = File::open(format!("{}/bzImage.config", blobs_path))
```
popegeo, line 715: the name of the config differs by platform

raulmldv: I changed the names to match the platform and also changed the blob names in samples/command_executer/resources/blobs

eif_defs/src/lib.rs
```diff
@@ -28,7 +28,8 @@ pub const EIF_HDR_ARCH_ARM64: u16 = 0x1;
 /// of this structures, we assume changes are backwards compatible.
 /// V0 -> V1: Add support to generate and check CRC.
 /// V1 -> V2: Add the signature section.
-pub const CURRENT_VERSION: u16 = 3;
+/// V2 -> V3: Add the metadata section.
+pub const CURRENT_VERSION: u16 = 4;
```
aatrand, line 32: This is confusing: The comments suggest that the current version should be V3, but the constant's value is 4.

eif_utils/src/lib.rs
costisin: Image.config?

costisin (commented):
lgtm

eif_utils/src/lib.rs
raulmldv: Yes, that's the new name.

eif_defs/src/lib.rs
raulmldv: I agree it's confusing so I changed all version transitions to start from V1 instead of V0.

enclave_build/src/lib.rs
```diff
@@ -73,6 +79,31 @@ impl<'a> Docker2Eif<'a> {
...
+        // TODO: set default values for image name and version in next commit
+        let default = &"default".to_string();
+        let img_name = match img_name {
+            Some(name) => name,
+            None => default,
+        };
+        let img_version = match img_version {
+            Some(version) => version,
+            None => default,
+        };
```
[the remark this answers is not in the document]
raulmldv: Added a constant.

costisin (approved):

eif_utils/src/lib.rs
```diff
@@ -519,15 +669,71 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
...
+            std::env::var("NITRO_CLI_SPECS").unwrap_or_else(|_| DEFAULT_SPECS_PATH.to_string());
+        let specs_file = File::open(format!("{}/aws-nitro-enclaves-cli.spec", specs_path))
+            .expect("Failed to open NitroCLI specs file.");
+        let spec_lines = BufReader::new(specs_file).lines();
+        let mut tool_name = String::new();
+        let mut tool_version = String::new();
+
+        for line in spec_lines {
+            let line = line.expect("Failed to extract line from specs");
+            let clean_line = line.replace(' ', "");
```
aatrand, line 705: Nit: This may not work as expected if the file contains tabs instead of blanks.

eif_utils/src/lib.rs
```diff
@@ -519,15 +669,71 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
...
+            let split_line: Vec<&str> = clean_line.split(':').collect();
+
+            if line.contains("Name:") {
+                tool_name = split_line[1].to_string();
```
aatrand, line 709: Nit: You should check that ```split_line``` has at least 2 elements, otherwise a malformed file may cause a crash here.

eif_utils/src/lib.rs
```diff
@@ -519,15 +669,71 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
...
+            .lines()
+            .nth(2)
+            .unwrap()
+            .expect("Failed to read kernel config file.");
+
+        // Extract OS and version from line format:
+        // ' # Linux/x86_64 4.14.177-104.253.amzn2.x86_64 Kernel Configuration '
+        let sep: Vec<char> = vec![' ', '/', '-'];
+        let os_words: Vec<&str> = os_string.split(&sep[..]).collect();
+        meta.insert("OperatingSystem".to_string(), os_words[1].to_string());
```
aatrand, line 731: Same observation as above: Check first that indices 1 and 3 are valid and consider using defaults if the tokenized string doesn't contain valid words at indices 1 or 3.

enclave_build/src/lib.rs
```diff
@@ -73,6 +83,33 @@ impl<'a> Docker2Eif<'a> {
...
+                    return Err(Docker2EifError::MetadataPathError);
+                } else {
+                    Some(meta.clone())
+                }
+            }
+            None => None,
+        };
+
+        let uri_split: Vec<&str> = docker_image.split(':').collect();
+        let repo = uri_split[0].to_string();
```
aatrand, line 98: What if ```uri_split``` has no elements? You should check this and return an error here.

aatrand (dismissed):

eif_utils/src/lib.rs
raulmldv: Added the suggested checks.

alexggh (changes requested):

alexggh (dismissed):

enclave_build/src/lib.rs
shtaked: Added check for `is_empty()`

enclave_build/src/lib.rs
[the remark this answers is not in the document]
shtaked: Not relevant anymore, docker image name and tag are used as defaults

eif_utils/src/lib.rs
shtaked: outdated

eif_utils/src/lib.rs
shtaked: outdated, `.spec` file is not parsed anymore

eif_utils/src/lib.rs
```diff
@@ -521,15 +671,58 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
...
 
+    pub fn get_generated_meta(&mut self) -> BTreeMap<String, String> {
+        let mut meta = BTreeMap::new();
+
+        let time = SystemTime::now();
+        let time: DateTime<Utc> = time.into();
+        meta.insert("BuildTime".to_string(), time.to_rfc3339());
+
+        // Get package version and name from CARGO_PKG_VERSION and CARGO_PKG_NAME
+        let tool_name = env!("CARGO_PKG_NAME").to_string();
```
shtaked, line 693: It puts the name and version of the current package: `eif_utils` while probably expected to put the root `nitro_cli` package details

Makefile
```diff
@@ -318,6 +319,8 @@ install-tools:
 	$(CP) -r blobs/${HOST_MACHINE}/* ${NITRO_CLI_INSTALL_DIR}${DATA_DIR}/nitro_enclaves/blobs/
 	$(MKDIR) -p ${NITRO_CLI_INSTALL_DIR}${DATA_DIR}/nitro_enclaves/examples
 	$(CP) -r examples/${HOST_MACHINE}/* ${NITRO_CLI_INSTALL_DIR}${DATA_DIR}/nitro_enclaves/examples/
+	$(MKDIR) -p ${NITRO_CLI_INSTALL_DIR}${DATA_DIR}/nitro_enclaves/SPECS
```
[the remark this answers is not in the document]
shtaked: Spec is not used anymore

eif_utils/src/lib.rs
```diff
@@ -521,15 +672,79 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
...
+        let time = SystemTime::now();
+        let time: DateTime<Utc> = time.into();
+        meta.insert("BuildTime".to_string(), time.to_rfc3339());
+
+        let blobs_path =
+            std::env::var("NITRO_CLI_BLOBS").unwrap_or_else(|_| DEFAULT_BLOBS_PATH.to_string());
+        // Read aws-nitro-enclaves-cli.spec file and parse the lines that contain:
+        // Name:      aws-nitro-enclaves-cli
+        // Version:   1.0.12
+        let specs_path =
```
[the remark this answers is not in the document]
shtaked: spec is not used anymore

eif_utils/src/lib.rs
shtaked: That's also the reason for unit tests failure

tests/tests.rs
```diff
@@ -708,6 +778,157 @@ mod tests {
...
+            &args.docker_dir,
+            &args.output,
+            &args.signing_certificate,
+            &args.private_key,
+            &args.img_name,
+            &args.img_version,
+            &args.metadata,
+        )
+        .expect("Docker build failed")
+        .1;
```
shtaked, line 822: `.1` is redundant here

alexggh (dismissed):

src/common/commands_parser.rs
```diff
@@ -207,10 +216,26 @@ impl ConsoleArgs {
...
 }
 
-/// The arguments used by the `describe-enclaves` command.
+/// Empty set of arguments.
 #[derive(Serialize, Deserialize)]
 pub struct EmptyArgs {}
 
+/// The arguments used by `describe-enclaves` command.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DescribeArgs {
```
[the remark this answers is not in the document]
shtaked: fixed

src/lib.rs
```diff
@@ -36,6 +37,8 @@ use log::info;
 
 use utils::{Console, PcrType};
 
+use crate::common::json_output::DescribeMetadata;
```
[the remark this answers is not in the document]
shtaked: fixed

src/common/json_output.rs
```diff
@@ -167,7 +186,7 @@ impl EnclaveBuildInfo {
 pub struct DescribeEifInfo {
```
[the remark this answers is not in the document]
shtaked: fixed

src/common/json_output.rs
```diff
@@ -178,21 +197,119 @@ pub struct DescribeEifInfo {
...
+            img_name,
+            img_version,
+            metadata,
         }
     }
 }
+
+/// Metadata to be included in the describe output
+#[derive(Clone, Serialize, Deserialize)]
+pub struct DescribeMetadata {
```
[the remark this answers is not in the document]
shtaked: done

eif_utils/src/lib.rs
shtaked: Moved to a separate `EifBuildInfo` struct which is propagated as argument. `generate_build_info!` macros can be used to generate it based on current cargo package

enclave_build/src/lib.rs
```diff
@@ -118,6 +168,17 @@ impl<'a> Docker2Eif<'a> {
     }
 
     pub fn create(&mut self) -> Result<BTreeMap<String, String>, Docker2EifError> {
+        let info = async {
```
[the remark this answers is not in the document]
shtaked: done

eif_utils/src/bin/eif_build.rs
```diff
@@ -122,6 +141,27 @@ fn main() {
         _ => panic!("Both signing-certificate and private-key parameters must be provided"),
     };
 
+    let img_name = matches.value_of("image_name").map(|val| val.to_string());
+    let img_version = matches.value_of("image_name").map(|val| val.to_string());
+    let metadata = matches.value_of("metadata").map(|val| val.to_string());
+
+    let arg_data = EifIdentityInfo {
```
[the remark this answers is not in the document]
shtaked: ranamed to `eif_info`

eif_utils/src/lib.rs
```diff
@@ -521,15 +671,58 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
...
+
+        // Get package version and name from CARGO_PKG_VERSION and CARGO_PKG_NAME
+        let tool_name = env!("CARGO_PKG_NAME").to_string();
+        let tool_version = env!("CARGO_PKG_VERSION").to_string();
+
+        meta.insert("BuildTool".to_string(), tool_name);
+        meta.insert("BuildToolVersion".to_string(), tool_version);
+
+        let blobs_path =
+            std::env::var("NITRO_CLI_BLOBS").unwrap_or_else(|_| DEFAULT_BLOBS_PATH.to_string());
```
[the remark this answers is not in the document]
shtaked: Passed explicitly internally now

eif_utils/src/lib.rs
```diff
@@ -131,6 +169,8 @@ pub struct EifBuilder<T: Digest + Debug + Write + Clone> {
     /// Hash the signing certificate
     pub certificate_hasher: EifHasher<T>,
     hasher_template: T,
+    /// Metadata of the EIF
+    eif_data: EifIdentityInfo,
```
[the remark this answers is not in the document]
shtaked: renamed to `eif_info`

src/enclave_proc/mod.rs
```diff
@@ -414,11 +428,10 @@ fn process_event_loop(
     let mut conn_listener = ConnectionListener::new()?;
     let mut enclave_manager = EnclaveManager::default();
     let mut terminate_thread: Option<std::thread::JoinHandle<()>> = None;
-    let mut pcr_thread: Option<std::thread::JoinHandle<NitroCliResult<BTreeMap<String, String>>>> =
-        None;
+    let mut pcr_thread: PcrThread = None;
     let mut done = false;
     let mut ret_value = Ok(());
-    let mut add_info = false;
+    let mut add_info = InfoLevel::Basic;
```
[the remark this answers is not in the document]
shtaked: removed completely

src/enclave_proc/mod.rs
```diff
@@ -358,41 +358,55 @@ fn handle_command(
...
                             "Failed to set measurements inside enclave handle.".to_string(),
                         )
                     })?;
+                enclave_manager.set_metadata(metadata).map_err(|e| {
+                    e.add_subaction("Failed to set metadata inside enclave handle.".to_string())
+                })?;
                 *pcr_thread = None;
             }
 
+            if describe_args.metadata {
```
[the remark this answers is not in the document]
shtaked: fixed

src/enclave_proc/commands.rs
```diff
@@ -5,28 +5,32 @@
...
 use crate::enclave_proc::connection::Connection;
 use crate::enclave_proc::connection::{safe_conn_eprintln, safe_conn_println};
 use crate::enclave_proc::cpu_info::CpuInfo;
 use crate::enclave_proc::resource_manager::{EnclaveManager, EnclaveState};
-use crate::enclave_proc::utils::get_enclave_describe_info;
+use crate::enclave_proc::utils::{get_enclave_describe_info, InfoLevel};
 use crate::new_nitro_cli_failure;
 
+/// Thread handle from parallel computing of PCRs
+pub type PcrThread = Option<JoinHandle<NitroCliResult<(BTreeMap<String, String>, Value)>>>;
```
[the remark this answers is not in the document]
shtaked: renamed to `DescribeThread`

src/enclave_proc/mod.rs
```diff
@@ -358,41 +358,55 @@ fn handle_command(
...
+                    .set_action("Describe Enclave".to_string())
+            })?;
             connection.write_u64(MSG_ENCLAVE_CONFIRM).map_err(|e| {
                 e.add_subaction("Failed to write confirmation".to_string())
                     .set_action("Describe Enclaves".to_string())
             })?;
 
             // Evaluate thread result at first describe, then set thread to None.
             if pcr_thread.is_some() {
-                enclave_manager
```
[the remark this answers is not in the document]
shtaked: Moved to `fetch_describe_result` function.

But generally I would go further and refactor the approach of `DescribeThread` logic: it is to implicit and prone to bugs. It would be better to keep the thread logic within `EnclaveManager` itself rather then setting the measurements and metadata result from the outside.

But probably such a change not for now, this MR is already too huge.

eif_utils/src/lib.rs
```diff
@@ -176,9 +219,12 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
         self.ramdisks.push(ramdisk_file);
     }
 
-    /// The first two sections are the kernel and the cmdline.
+    /// The first two sections are the kernel and the cmdline and the last is metadata.
```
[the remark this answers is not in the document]
shtaked: done

enclave_build/src/lib.rs
```diff
@@ -58,6 +62,9 @@ impl<'a> Docker2Eif<'a> {
         artifacts_prefix: String,
         certificate_path: &Option<String>,
         key_path: &Option<String>,
+        img_name: &Option<String>,
+        img_version: &Option<String>,
+        meta_file_path: &Option<String>,
```
petreeftime: These could be put into a separate structure. This structure should implement Default so that we can get rid of the unwraps lower in the function.

enclave_build/src/lib.rs
```diff
@@ -73,6 +80,31 @@ impl<'a> Docker2Eif<'a> {
...
+                        Docker2EifError::MetadataFileError
+                    })?)
+                }
+            }
+            None => None,
+        };
+
+        // TODO: set default values for image name and version in next commit
+        let default = &"default".to_string();
+        let img_name = match img_name {
```
petreeftime, line 106: `img_name.unwrap_or_else`

src/lib.rs
```diff
@@ -702,6 +711,24 @@ macro_rules! create_app {
                             .long("private-key")
                             .help("Local path to developer's Eliptic Curve private key.")
                             .takes_value(true),
+                    )
+                    .arg(
+                        Arg::with_name("image_name")
```
petreeftime, line 744: Add a `default` value here as well.

eif_utils/src/bin/eif_build.rs
```diff
@@ -122,6 +141,27 @@ fn main() {
...
+        img_name: match img_name {
+            Some(name) => name,
+            // Set default value kernel file name
+            None => {
+                let path_split: Vec<&str> = kernel_path.split('/').collect();
+                path_split[path_split.len() - 1].to_string()
```
petreeftime, line 174: What does this do? If it tries to get the last bit, use `Path` and get the `file_name`.

enclave_build/src/lib.rs
```diff
@@ -12,10 +12,14 @@ mod yaml_generator;
 use docker::DockerUtil;
 use eif_defs::EIF_HDR_ARCH_ARM64;
 use eif_utils::{EifBuilder, EifIdentityInfo, SignEnclaveInfo};
+use serde_json::json;
 use sha2::Digest;
 use std::collections::BTreeMap;
+use tokio::runtime::Runtime;
 use yaml_generator::YamlGenerator;
 
+pub const DEFAULT_TAG: &str = "1.0";
```
petreeftime, line 21: The default tag is usually `latest`, which doesn't make a lot of sense in this case. I suggest using the SHA256 value of the image?

Actually, better use the SHA256 value anyway, since tags are mutable.

eif_defs/src/lib.rs
```diff
@@ -26,9 +26,10 @@ pub const EIF_HDR_ARCH_ARM64: u16 = 0x1;
 
 /// Current EIF version to be incremented every time we change the format
 /// of this structures, we assume changes are backwards compatible.
-/// V0 -> V1: Add support to generate and check CRC.
-/// V1 -> V2: Add the signature section.
-pub const CURRENT_VERSION: u16 = 3;
+/// V1 -> V2: Add support to generate and check CRC.
+/// V2 -> V3: Add the signature section.
+/// V3 -> V4: Add the metadata section.
+pub const CURRENT_VERSION: u16 = 4;
```
petreeftime, line 31: Does the Enclave device expose the supported EIF version? Should nitro-cli try to make images compatible with older versions if the new features aren't required?

eif_utils/src/lib.rs
```diff
@@ -257,6 +237,10 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
 
         if self.sign_info.is_some() {
             result[DEFAULT_SECTIONS_COUNT as usize + self.ramdisks.len()] = self.signature_offset();
+            result[DEFAULT_SECTIONS_COUNT as usize + self.ramdisks.len() + 1] =
```
petreeftime: Should add a comment explaining this line.

eif_utils/src/lib.rs
```diff
@@ -382,6 +382,59 @@ impl<T: Digest + Debug + Write + Clone> EifBuilder<T> {
         }
     }
 
+    /// Get metadata parts, add them to the structure and
+    /// serialize all the information as JSON
```
petreeftime: Is the JSON serialized in a reproducible manner? Fields should be sorted alphabetically, etc.

src/common/json_output.rs
```diff
@@ -185,17 +204,32 @@ pub struct DescribeEifInfo {
...
+    #[serde(rename = "Metadata")]
+    /// EIF metadata
+    pub metadata: Option<DescribeMetadata>,
 }
 
 impl DescribeEifInfo {
     /// Create describe information structure for EIF.
     pub fn new(
-        version: u16,
+        version: String,
```
petreeftime: I'm not sure why this change is here.

petreeftime (changes requested):

eif_utils/src/lib.rs
shtaked: Removed this piece in the last commit: metadata section was put higher (after cmdline and before ram disks sections)

eif_utils/src/lib.rs
shtaked: It depends on the implementation of `serde_json`. I guess it should be reproducible, but I need to check.

src/common/json_output.rs
shtaked: True, looks like no big reason to change this type

eif_defs/src/lib.rs
shtaked: No, looks like there is no any exposal of supported EIF versions yet.
Specifically for metadata section VMM checks internally the version of loaded eif and extracts the metadata only if the version >= 4. As an alternative we could make metadata section fully optional even for new EIF version.
But, indeed, better to think in the future about some compatibility check mechanism.

enclave_build/src/lib.rs
shtaked: `img_name` default value depends on one other argument - `docker_image`.
As agreed offline, would be better to combine all arguments of that function to some common struct, i.e. `Docker2EifArgs`. But that should be addressed in a separate PR.

src/lib.rs
shtaked: Do you mean add a description of what will be taken as a default?
If you mean the value of the argument itself, then it's `Optional<String>` and will be propagated this way up to `Docker2Eif::new` where in case of `None` will be replaced with docker tag (or hash as discussed somewhere here)

eif_utils/src/lib.rs
shtaked: From what I understood, by default `serde_json` sorts the fields alphabetically as you said, until you import it with `preserve_order` feature enabled. That only works for `serde:json::Value` objects as for structs they are represented as a `Map` of specific type (depending on `preserve_order` flag).
Thus, if we have serializable `let eif_info EifIdentityInfo` struct, it can be serialized with alphabetical order of fields the following way:
```
serde_json::to_vec(serde_json::to_value(eif_info))
```
While we currently have it like this
```
serde_json::to_vec(eif_info)
```
and it will **not** sort the fields alphabetically.

The question here, is it important for metadata to have a specific order considering it's not affecting the PCR values?

src/common/json_output.rs
shtaked: reverted

enclave_build/src/lib.rs
shtaked: done

eif_utils/src/bin/eif_build.rs
shtaked: fixed

enclave_build/src/lib.rs
shtaked: done

petreeftime (approved):

aatrand (approved):

PR #302, opened 2021-09-10
Update aarch64 blobs

popegeo:
*Issue #, if available:*

*Description of changes:*


andraprs (approved):

src/common/mod.rs
```diff
@@ -568,7 +572,7 @@ mod tests {
 
     fn unset_envvar(varname: &String) {
         let _ = unsafe {
-            libc::unsetenv(varname.as_ptr() as *const i8);
+            libc::unsetenv(varname.as_ptr() as *const c_char);
```
costisin, line 575: why is this change needed? 

```
pub type c_char = i8;
```
[https://doc.rust-lang.org/std/os/raw/type.c_char.html](https://doc.rust-lang.org/std/os/raw/type.c_char.html)

src/lib.rs
```diff
@@ -101,11 +101,18 @@ pub fn build_from_docker(
             .add_info(vec![output_path, "Open"])
         })?;
 
+    let mut kernel_image_name = "bzImage";
+    if std::env::consts::ARCH.eq("aarch64") {
+        kernel_image_name = "Image";
+    } else if std::env::consts::ARCH.eq("x86_64") {
+        kernel_image_name = "bzImage";
+    }
```
costisin, line 109:
```suggestion
    let kernel_image_name = match std::env::consts::ARCH.as_str() {
        "aarch64" => "Image",
        "x86_64" => "bzImage",
    }
```

src/common/mod.rs
popegeo: If you read the text below:
This type will always be either i8 or u8, as the type is defined as being one byte long.

popegeo: https://github.com/rust-lang/rust/issues/79089

src/lib.rs
popegeo: that's a fancy suggestion, thanks

src/common/mod.rs
costisin: I got it!

andraprs (approved):

laurstefan (approved):

PR #310, opened 2021-10-05
Enclave boot timeout calculation

dblanzeanu:
*Issue #, if available:*

*Description of changes:*
- Add allow deref_nullptr so that the code can be compiled with stable version of Rust(1.54)
- Modify the nitro cli poll timeout to be calculated based on the EIF size used to create the Enclave.


src/enclave_proc/resource_manager.rs
```diff
@@ -1164,3 +1176,69 @@ pub fn between_packets_delay() -> Option<Duration> {
...
+/// let timeout = calculate_necessary_timeout(8 * GiB, 0);
+/// ```
+fn calculate_necessary_timeout(eif_size: u64, allocated_memory_mib: u64) -> c_int {
+    let poll_timeout: c_int;
+
+    if eif_size > 0 {
+        // in case we have a valid eif_size give 1.5 minutes for each 6GiB
+        poll_timeout = ((1 + (eif_size - 1) / (6 * GiB)) as i32).saturating_mul(TIMEOUT_MINUTE_MS);
+    } else {
+        // Update the poll timeout to be 1 minute per 100 GiB of enclave memory.
```
bveaws, line 1205: Why 1 min per 100 GB? was this measured somehow?

We recommend to customers that the enclave memory is 4 times the EIF size, minimum. This means that the EIF size here could be 25 GB.

dblanzeanu: This was the previous implementation and it did not take into account the EIF size.

dblanzeanu: I eliminated the legacy way of calculating the timeout.

PR #336, opened 2021-12-28
Add option to run-enclave to attach console

jethrogb:
Sometimes, if the enclave terminates very quickly, it can be difficult to attach the console quickly enough. This adds an option to run-enclave to attach as soon as possible.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the Apache License, version 2.0.


src/common/commands_parser.rs
```diff
@@ -449,6 +452,16 @@ fn debug_mode(args: &ArgMatches) -> Option<bool> {
     }
 }
 
+/// Parse the debug-mode flag from the command-line arguments.
```
petreeftime, line 455: `Parse the attach-console ...`.

src/common/commands_parser.rs
```diff
@@ -449,6 +452,16 @@ fn debug_mode(args: &ArgMatches) -> Option<bool> {
...
+fn attach_console(args: &ArgMatches) -> Option<bool> {
```
petreeftime, line 456: I don't really understand why Option<bool> was used for `debug-mode`, but should just be `bool`.

src/main.rs
```diff
@@ -114,12 +114,20 @@ fn main() {
...
             replies.push(comm);
-            enclave_process_handle_all_replies::<EnclaveRunInfo>(&mut replies, 0, false, vec![0])
+            let run_info = enclave_process_handle_all_replies::<EnclaveRunInfo>(&mut replies, 0, false, vec![0])
                 .map_err(|e| {
                     e.add_subaction("Failed to handle all enclave process replies".to_string())
                         .set_action(RUN_ENCLAVE_STR.to_string())
                 })
                 .ok_or_exit_with_errno(None);
+            if run_args.attach_console.unwrap_or(false) {
+                console_enclaves(run_info[0].enclave_cid, None)
```
petreeftime, line 124: There is an assumption here is that there's 1 enclave process which contains a valid CID, which I think will always be true, but I'm not sure that if there's more than 1 enclave process, the one found at index 0 will be the one that was just started. @aatrand what do you think?

petreeftime:
Thank you for the PR. It's definitely useful functionality!

src/main.rs
jethrogb: The array elements in `run_info` map 1:1 to the elements in `replies`. `replies` is initialized in this function with exactly 1 element.

petreeftime:
I've opened #340 so that I can fixup the above changes and make the tests happy. I'll close this one and focus on merging #340.

jethrogb:
Whoops I made the change but forgot to push.

PR #340, opened 2022-01-14
Auto attach console

petreeftime:
*Issue #, if available:*
https://github.com/aws/aws-nitro-enclaves-cli/pull/336

*Description of changes:*
Add console auto attach on run.


src/main.rs
```diff
@@ -114,12 +114,25 @@ fn main() {
...
+                false,
+                vec![0],
+            )
+            .map_err(|e| {
+                e.add_subaction("Failed to handle all enclave process replies".to_string())
+                    .set_action(RUN_ENCLAVE_STR.to_string())
+            })
+            .ok_or_exit_with_errno(None);
+            if run_args.attach_console {
+                console_enclaves(run_info[0].enclave_cid, None)
```
aatrand, line 129: Nit: we are theoretically guaranteed at this point that `run_info[0]` exists and is a valid `EnclaveRunInfo` object. But I would also add a check for this, for good measure.

aatrand (approved):

alexggh (approved):

petreeftime:
I rebased, tested and pushed directly based on the feedback.

PR #370, opened 2022-04-25
Create driver-bindings crate with static bindings

sabin-rapan:
The purpose of this refactoring stems from the complexities added by
bindgen's cc crate when cross-compiling nitro-cli:

When (cross)compiling nitro-cli for different {arch, toolchain} tuples
we need to be extra careful to pass the right sysroot to bindgen's cc
crate (see builds.rs: get_gcc_sysroot()). At the moment,
get_gcc_sysroot() is incorrect because it asks the cross-compiler where
its compiler sysroot is (by calling *-gcc --print-sysroot), which is not
necessarily the correct sysroot for your target, e.g. if you have
x86_64-unknown-linux-gnu-gcc installed in path A with sysroot in
${A}/sys-root and want to cross-compile for x86_64-unknown-linux-musl
where the sysroot is in ${B}/sys-root, you are incorrectly compiling for
a different toolchain. Even in the most fortunate case where everything
works properly (as it has done so far), we are are the mercy of the
Linux distro and/or sysadmin installing the toolchain packages.

Additionally, the benefit of dynamically generating driver bindings is
marginal at best (the APIs exposed by
linux/include/uapi/linux/nitro_enclaves.h are 3 structs and a couple of
defines).

Lastly, the userspace API of nitro_enclaves.h is very unlikely to change
in a backwards-incompatible manner; it's uapi after all.

Therefore, we choose to generate the bindings statically, commit them to
this repo and distribute them as another crate in the workspace. This
removes all the complexities of bindgen, allows us to precisely pinpoint
specific kernel versions supported by these bindings and opens the door
to split this crate to its own repo in the future so we can publish it
on crates.io and allow others to write their own rust applications that
talk to /dev/nitro_enclaves.

Unfortunately, build.rs also set the COMMIT_ID env var that was used
when reporting errors for git tree builds. Now that must go away. It
didn't provide much value anyway: if you build nitro-cli from github you
surely know the commit id you built it from. Instead we print the
version of the package.

Signed-off-by: Sabin Rapan <sabrapan@amazon.com>

*Issue #, if available:*

*Description of changes:*


driver-bindings/src/wrappers.rs
```diff
@@ -0,0 +1,13 @@
...
+use crate::bindings::ne_enclave_start_info;
+
+impl Default for ne_enclave_start_info {
+    fn default() -> Self {
+        Self {
+            flags: 0,
+            enclave_cid: 0,
+        }
+    }
+}
```
shtaked, line 13: nit: May be also move here the aliases from `resource_manager.rs` if you are going to move it to separate repo later?
```
/// CamelCase alias for the bindgen generated driver struct (ne_enclave_start_info).
pub type EnclaveStartInfo = ne_enclave_start_info;

/// CamelCase alias for the bindgen generated driver struct (ne_user_memory_region).
pub type UserMemoryRegion = ne_user_memory_region;

/// CamelCase alias for the bindgen generate struct (ne_image_load_info).
pub type ImageLoadInfo = ne_image_load_info;
```

shtaked (approved):

driver-bindings/src/wrappers.rs
sabin-rapan: I thought about this a lot myself. It's a really good pattern to provide sane wrappers over plain FFI bindings (e.g. https://github.com/rust-vmm/kvm-bindings/blob/main/src/x86/fam_wrappers.rs), and my first attempt was to move these inside the bindings crate, but then I realized we don't add any APIs to them whatsoever. More so, we use them as C structs (a few lines below the snippet you showed we size_of() on one of these structs). In their current state, these aliases wouldn't provide any real value if they were apart of the bindings crate.

If you're ok with it, we can move them if/when their API evolves beyond their current impl.

shtaked: Yep, I'm ok with that, my only idea was to at least bring naming style closer to Rust. But it's fine for now.

petreeftime (approved):

PR #371, opened 2022-05-17
Update chrono dependency

Pleto:
*Issue #, if available:* None

*Description of changes:*
Chrono package is no longer maintained and it's marked as having a security vulnerability. I replaced it with [time](https://docs.rs/time/latest/time/) since it's still active and offers the same API as chrono

### DIFF
--- a/drivers/virt/nitro_enclaves/ne_misc_dev.c
+++ b/drivers/virt/nitro_enclaves/ne_misc_dev.c
@@ -1844,35 +1844,10 @@ static long ne_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 
 #if defined(CONFIG_NITRO_ENCLAVES_MISC_DEV_TEST)
 #include "ne_misc_dev_test.c"
-
-static inline int ne_misc_dev_test_init(void)
-{
-	return __kunit_test_suites_init(ne_misc_dev_test_suites);
-}
-
-static inline void ne_misc_dev_test_exit(void)
-{
-	__kunit_test_suites_exit(ne_misc_dev_test_suites);
-}
-#else
-static inline int ne_misc_dev_test_init(void)
-{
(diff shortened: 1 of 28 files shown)


drivers/virt/nitro_enclaves/ne_misc_dev.c
```diff
@@ -140,7 +140,7 @@ static int user_access_ok(void __user *addr, unsigned long size)
 #endif
 }
 
-#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 7, 0)
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 4, 0)
```
andraprs, line 143: These kernel version markers are set to follow the upstream Linux kernel, as in which version from upstream, the functionality has been included. The main source for the Nitro Enclaves kernel driver is the upstream Linux kernel.

If still needed, this out-of-tree codebase from the drivers folder could be used for the systems that do not have the NE kernel driver already integrated. Amazon Linux 2 already has the NE kernel driver included in their kernels. One option could be to comment / remove these lines, if needed to test / debug using this source code. 

Open to discuss if certain use cases would need to consider other options for this out-of-tree codebase.

Pleto: Ok, got it. 

> One option could be to comment / remove these lines, if needed to test / debug using this source code.

I initially removed these lines to test locally, but found it a bit weird, so decided to change the required kernel version. I'll drop this commit. Thanks for the explanation!

petreeftime:
Please don't include tags such as `[chore]` in the commit name. Also, try to make commits atomic. If you update Cargo.toml, you also need to update Cargo.lock and the license files in the same commit.

petreeftime (dismissed):

Issue #235, opened 2021-03-06
Build enclave image without Docker daemon from local docker achive


sphw:
### What
Right now, you need a running Docker daemon to build an enclave image. The LinuxKit version currently included attempts to pull the image using the Docker daemon. https://github.com/linuxkit/linuxkit/pull/3573 now lets LinuxKit pull the images directly without the need for a Docker daemon. 

### Why
As part of our enclave support, we ([M10 Networks](https://m10.io)) want to be able to build enclave images entirely in a Dockerfile. We distribute all of our services through Docker images, and all of the builds are entirely performed inside of Dockerfiles. While possible to change this for Nitro, I don't think it should be necessary with a simple change. 

### How
To supply this functionality, more or less out of the box, all that would be required would be to update the included LinuxKit to one of the latest builds. This would then allow users to use [skopeo](https://github.com/containers/skopeo), or a similar tool, to transfer their local image into LinuxKit's cache located at `~/.linuxkit/cache` themselves. 

A more user-friendly follow on would be to allow the users to simply pass in a path to their own Docker image archive. At that point, the CLI would need to copy the image into the LinuxKit cache. 

As a temporary workaround for our use case, I can simply replace the LinuxKit binary in `/usr/share/nitro_enclaves/blobs` with my own. I think this use case is common enough that it should either be supported through easier means or documented in some way

sphw:
On further investigation there is actually a further reason the `nitro-cli` calls out to the Docker daemon to inspect the Docker image: https://github.com/aws/aws-nitro-enclaves-cli/blob/0e164db23194d9f3bca94c8f5ba42f869fbcdc7c/enclave_build/src/docker.rs#L252 

There are a few ways this could be fixed. It might make sense to allow users to simply configure the env vars, and cmd themselves. That would be a useful as a feature in its own right. I am happy to open a PR adding those options to the CLI if there is interest.

petreeftime:
Our LinuxKit also has an additional feature which doesn't seem to be merged: https://github.com/linuxkit/linuxkit/pull/3446 and needs to be rebased. @neacsu do you know what is missing for upstreaming the feature? Are there any changes we would need to do in the codebase to support it?

sphw:
> Our LinuxKit also has an additional feature which doesn't seem to be merged: linuxkit/linuxkit#3446 and needs to be rebased. @neacsu do you know what is missing for upstreaming the feature? Are there any changes we would need to do in the codebase to support it?

[...]

PR #373, opened 2022-05-23
linuxkit: Update linuxkit blobs to v0.8+

shtaked:
*Issue #, if available:*
https://github.com/aws/aws-nitro-enclaves-cli/issues/235

*Description of changes:*
Linuxkit blobs are updated from the upstream branch
with commit `b710224cdf9a8425a7129cdcb84fc1af00f926d7`

The blob is slightly modified to support prefixing the
filesystem of the docker image. This is necessary for
generating the initrd image.

To avoid size limitations present in the linux kernel,
the gzip wrapper is removed.

### DIFF
--- a/enclave_build/src/lib.rs
+++ b/enclave_build/src/lib.rs
@@ -240,6 +240,7 @@ impl<'a> Docker2Eif<'a> {
         let output = Command::new(&self.linuxkit_path)
             .args(&[
                 "build",
+                "-docker",
                 "-name",
                 &customer_ramfs,
                 "-format",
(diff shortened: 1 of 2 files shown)


enclave_build/src/lib.rs
```diff
@@ -240,6 +240,7 @@ impl<'a> Docker2Eif<'a> {
         let output = Command::new(&self.linuxkit_path)
             .args(&[
                 "build",
+                "-docker",
```
shtaked, line 243: New flag was added to first get the image from docker cache before linuxkit cache

petreeftime (approved):

popegeo (approved):

PR #375, opened 2022-05-25
nitro-cli: Fix race condition on command dispatch

Pleto:
Signed-off-by: Cosmin-Andrei Pletosu <pletosuc@amazon.com>

*Issue #, if available:* None

*Description of changes:*

When the CLI sends a command to all enclaves, it opens a socket with
each enclave process, sends the command to each open socket and waits
for a confirmation message from each enclave, marking that the enclave
is still alive and the command has been recieved. Afterwands, the CLI
starts consuming what's available in each socket's buffer
(namely, command's reply from each enclave).

The CLI waits for the confirmations using an epoll that monitors each
socket. Currently, after the confirmation has been recieved, the epoll
still waits for messages from the "confirmed" enclave, which can lead
to the circumstance in which the epoll is triggered by a socket event
that contains command's reply (from the "confirmed" enclave), and not
another confirmation message from another process. Since we're
expecting exactly "n" confirmation messages, this leaves the operation
in an invalid state. Deleting the socket fd after the first received
message (which should be a confirmation) fixes this issue.

Alongside this fix, I have also removed an unnecessary clone of the
aforesaid socket


src/enclave_proc_comm.rs
```diff
@@ -221,14 +215,29 @@ where
             continue;
         }
 
+        let input_stream_raw_fd = events[0].data() as RawFd;
+        let mut input_stream = unsafe { UnixStream::from_raw_fd(input_stream_raw_fd) };
```
petreeftime, line 219: Why is this safe?

Pleto: This is code that already exists in mainline. However, imo, this is not completely safe since the stream socket peer can close connection (or shut down), making this file descriptor invalid. I'll add an extra check on the EpollEvent received. Thanks!

Pleto: Initially I thought that the file descriptor will be invalidated if the epoll_wait returns an event with error flags set, but apparently it won't be, so this operation is safe

shtaked (approved):

src/enclave_proc_comm.rs
```diff
@@ -157,32 +158,25 @@ where
         .map_err(|e| {
             e.add_subaction("Failed to send command to all enclave processes".to_string())
         })?
-        .iter_mut()
-        .map(|socket| {
+        .into_iter()
+        .map(|mut socket| {
+            // Send the command.
+            enclave_proc_command_send_single(cmd, args, socket.borrow_mut())?;
```
sabin-rapan, line 164: It's more logical if the code is structured as: epoll ctl add, followed by enclave_proc_command_send_single, so that the thing that waits for events is created before the thing that generates the events

src/enclave_proc_comm.rs
```diff
@@ -221,14 +215,29 @@ where
...
+            EpollOp::EpollCtlDel,
+            input_stream_raw_fd,
+            Option::None,
+        )
+        .map_err(|e| {
+            new_nitro_cli_failure!(
+                &format!("Failed to remove socket from epoll: {:?}", e),
+                NitroCliErrorEnum::EpollError
+            )
+        })?;
```
sabin-rapan, line 240: Shouldn't this be in
```rust
           if reply == MSG_ENCLAVE_CONFIRM {
                debug!("Got confirmation from {:?}", input_stream);
                replies.push(input_stream);
            }
```
so that only enclave processes that confirmed the command are removed from the epoll loop? Otherwise, you would remove enclaves that either return an error in read_u64_le or a reply different from MSG_ENCLAVE_CONFIRM

src/enclave_proc_comm.rs
Pleto: I agree, but maintaining the logical order clutters the code. Any interaction with epoll is done with a RawFd, so we need to be extra careful to not close the socket (done automatically when the Socket object is dropped) and invalidate its fd. To keep the order you mentioned, I see three viable options:
- copy the socket and feed the epoll with the clone (this version is currently in mainline)
- have the map return socket's fd (and consume the socket): each closure ends with Ok(socket.into_raw_fd())
  - definitely the best solution, but some extra comments are needed for this line to explain why this is necessary + the result is never actually used / needed
- have the map return all sockets: each closure ends with Ok(socket). In this case, we also need to add some extra logic to guarantee that these sockets are used when we wait on epoll

I honestly prefer the version written in this commit, but if you believe otherwise, I can implement one of these options

src/enclave_proc_comm.rs
Pleto: No, the mini-protocol expects that the first message should always be "MSG_ENCLAVE_CONFIRM". If something else comes from the connection, we shouldn't wait again on the same socket in case it decides to eventually send a "MSG_ENCLAVE_CONFIRM": if the first message is wrong, then definitely something bad has happened with that enclave, so we should treat it as an error

andraprs (approved):

sabin-rapan (approved):

PR #383, opened 2022-06-16
Refactor console disconnect timeout feature

costinsin:
*Description of changes:*
Replace clutered timeout code with a TimerFd added
to epoll.

TimerFd is armed with `disconnect_timeout` seconds
and when the timer fires, the reading loop breakes.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -12,10 +12,12 @@ use nix::sys::time::{TimeVal, TimeValLike};
 use nix::unistd::read;
 use std::io::Write;
 use std::mem::size_of;
+use std::os::unix::io::IntoRawFd;
 use std::os::unix::io::RawFd;
 use std::thread::sleep;
-use std::time::{Duration, Instant, SystemTime};
+use std::time::{Duration, SystemTime};
 use vmm_sys_util::epoll::{ControlOperation, Epoll, EpollEvent, EventSet};
+use vmm_sys_util::timerfd::TimerFd;
 
 use crate::common::{NitroCliErrorEnum, NitroCliFailure, NitroCliResult};
 use crate::new_nitro_cli_failure;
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -146,54 +148,82 @@ impl Console {
...
+            // Wait for kernel notification that one of the fds is available
+            let num_events = epoll.wait(-1, &mut events).map_err(|e| {
+                new_nitro_cli_failure!(
+                    &format!("Failed to wait epoll: {:?}", e),
+                    NitroCliErrorEnum::EpollError
+                )
+            })?;
+
+            // Check if timerfd has triggered
+            if num_events > 1 || !events.map(|event| event.fd()).contains(&self.fd) {
```
shtaked, line 222: What is the meaning of first `if num_events > 1` check? Couldn't you keep just the second part alone?

src/utils.rs
```diff
@@ -146,54 +148,82 @@ impl Console {
...
             Epoll::new().map_err(|e| {
                 new_nitro_cli_failure!(
                     &format!("Failed to create epoll: {:?}", e),
                     NitroCliErrorEnum::EpollError
                 )
             })?,
-            [EpollEvent::default(); 1],
-            0,
-            Instant::now(),
+            [EpollEvent::default(); 2],
```
shtaked, line 158: I suppose we can keep expectation of a single `epoll` event and further avoid checking multiple events received.

shtaked, line 158: I suppose we can keep expectation of a single `epoll` event and further avoid checking multiple events received.

shtaked (changes requested):
Build fails with the following error:
```
error[E0658]: use of unstable library feature 'array_map'
   --> src/utils.rs:222:42
    |
222 |             if num_events > 1 || !events.map(|event| event.fd()).contains(&self.fd) {
    |                                          ^^^
    |
    = note: see issue #75243 <https://github.com/rust-lang/rust/issues/75243> for more information
```

src/utils.rs
costinsin: The first check is much more faster than the second check and will be true 99% of the time. I have introduced it to make the program more performant. The second check is slower because it converts it to an iterator and then searches through it, while the other translates into just a `cmp` assembly instruction.

If you want I can only keep the second check.

src/utils.rs
costinsin: I have changed this to only wait for one `epoll` event at a time. Now the program checks for the index 0 in the `events` array.

src/utils.rs
costinsin: Reworked!

shtaked (approved):

src/utils.rs
```diff
@@ -146,54 +148,84 @@ impl Console {
         disconnect_timeout_sec: Option<u64>,
     ) -> NitroCliResult<()> {
         // Initialize variables
-        let (epoll, mut events, mut epoll_timeout_us, mut start_epoll_time) = (
+        let (epoll, mut events) = (
             Epoll::new().map_err(|e| {
```
petreeftime, line 152: I can't say I understand why these are declared in a tuple. Seems pretty weird, especially since that tuple is not used anywhere else; they should just be declared normally, one after the other.

src/utils.rs
```diff
@@ -146,54 +148,84 @@ impl Console {
...
-                            NitroCliErrorEnum::EpollError
-                        )
-                    })?;
-
-                // If the timeout expires, no event happend and the console disconnects
-                if num_events == 0 {
+            // Wait for kernel notification that one of the fds is available
+            epoll.wait(-1, &mut events).map_err(|e| {
+                new_nitro_cli_failure!(
+                    &format!("Failed to wait epoll: {:?}", e),
```
petreeftime, line 216: This doesn't seem to properly deal with EINTR/EAGAIN errors. On EINTR/EAGAIN, the operation should be retried. We already have a timeout, so if the timeout is still active, we should not exit here.

src/utils.rs
```diff
@@ -146,54 +148,84 @@ impl Console {
...
-                        new_nitro_cli_failure!(
-                            &format!("Failed to wait on epoll: {:?}", e),
-                            NitroCliErrorEnum::EpollError
-                        )
-                    })?;
-
-                // If the timeout expires, no event happend and the console disconnects
-                if num_events == 0 {
+            // Wait for kernel notification that one of the fds is available
+            epoll.wait(-1, &mut events).map_err(|e| {
```
petreeftime, line 214: I don't really see that this is simpler to understand, both timeout and a timer_fd provide the same functionality, but timer_fd needs to be declared and armed, while timeout abstracts this. There were some improvements to be done to how things were handled previously, for sure, but I don't understand how the timerfd makes things easier to understand this case, and the commit message doesn't really clarify it for me either.

petreeftime (changes requested):

src/utils.rs
shtaked: I see a small benefit in eliminating the logic of manually decreasing the value of timer in a loop.

src/utils.rs
costinsin: I have changed the variable declaration to no longer be a tuple.

src/utils.rs
costinsin: Thanks for the observation! I have changed the code in a manner that if an interrupt unblocks the `wait` call and the triggered event number is `0`, the loop will just simply continue and block at the next `wait` call.

src/utils.rs
costinsin: I think that letting the kernel to manage the timer and notify the process when the timer expired is better than managing the timer in user space. Handling different corner cases with the math done for keeping track of elapsed time might be difficult and produce cluttered code.

Also, this version is more scalable, if you ever want to add another fd to epoll (eg. EventFd).

shtaked (approved):

petreeftime (approved):

PR #396, opened 2022-07-26
rpm install: Add integration test for env checking after RPM install

meerd:
**NOTE: Merging Gabriel's lingering commit.**

Added a separate integration test which checks if various components
are configured correctly (vsock-proxy service, hugetlbfs and CPU pool)
after installing the Nitro CLI package.


andraprs (approved):

tests/integration/test_setup.py
```diff
@@ -0,0 +1,134 @@
...
+
+    assert status == 0, 'vsock-proxy service is not running'
+
+def test_hugetlbfs_isset():
+    """
+    Tests that after the oneshot configuration service has been triggered
+    (at least once, after the package installation), at least one NUMA
+    node has some hugepages (any size) allocated
+
+    Sample path which is to be checed:
```
sabin-rapan, line 85: s/checed/checked

tests/integration/test_setup.py
```diff
@@ -0,0 +1,134 @@
...
+
+def test_vsock_proxy_is_running():
+    """
+    Check that the vsock-proxy service is running as soon as
+    the package is installed
+    """
+
+    info = 'Main PID: [0-9]+'
+
+    status = os.system('systemctl status vsock-proxy.service > /dev/null 2>&1')
```
sabin-rapan, line 67: Why check twice?

sabin-rapan (changes requested):

PR #404, opened 2022-08-23
Dockerless nitro-cli

calincoman:
*Issue #, if available:*

*Description of changes:*
- image.rs: contains the ImageDetails struct, which represents the metadata of an image in OCI-compliant format.
- pull.rs: contains logic for pulling an image from a remote registry, as well as authentication.
- cache.rs: contains the implementation of the local filesystem cache, providing methods for storing and fetching images
                   to/from cache.
- image_manager.rs: ImageManager struct, wrapper for the cache implementation.
- Refactored docker.rs to use the new code instead of the shiplift library.
- Updated Rust version in the tools/Dockerfiles to 1.59.0 in order to support a language feature used by a dependency.

### DIFF
--- a/enclave_build/Cargo.toml
+++ b/enclave_build/Cargo.toml
@@ -3,7 +3,7 @@ name = "enclave_build"
 version = "0.1.0"
 authors = ["The AWS Nitro Enclaves Team <aws-nitro-enclaves-devel@amazon.com>"]
 edition = "2018"
-rust-version = "1.58"
+rust-version = "1.59"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
@@ -20,5 +20,8 @@ log = "0.4"
 url = "2.1"
 sha2 = "0.9.5"
 futures = "0.3.13"
+thiserror = "1.0"
 
 aws-nitro-enclaves-image-format = "0.2"
+oci-distribution = "0.9.2"
+oci-spec = "0.5.8"
--- a/enclave_build/src/cache.rs
+++ b/enclave_build/src/cache.rs
@@ -1,0 +1,734 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// S&PDX-License-Identifier: Apache-2.0
+
+use std::{
+    collections::HashMap,
+    fs::{self, File, OpenOptions},
+    io::{Error, ErrorKind, Read, Write},
+    path::{Path, PathBuf},
+};
+
+use sha2::Digest;
+
+use crate::{
+    image::{self, ImageDetails},
+    EnclaveBuildError, Result,
+};
+
+use oci_distribution::client::ImageData;
+
+/// Root folder for the cache.
+pub const CACHE_ROOT_FOLDER: &str = "XDG_DATA_HOME";
+/// Name of the cache index file which stores the (image URI <-> image hash) mappings.
+pub const CACHE_INDEX_FILE_NAME: &str = "index.json";
+/// The name of the folder used by the cache to store the image layers.
+pub const CACHE_LAYERS_FOLDER_NAME: &str = "layers";
+/// The name of the image config file from the cache.
+pub const CACHE_CONFIG_FILE_NAME: &str = "config.json";
+/// The name of the image manifest file from the cache.
+pub const CACHE_MANIFEST_FILE_NAME: &str = "manifest.json";
+
+/// Struct which provides operations with the local cache.
+///
+/// The index.json file is located in the cache root folder and keeps track of images stored in cache.
+///
+/// The cache structure is:
+///
+/// {CACHE_ROOT_PATH}/index.json\
+/// {CACHE_ROOT_PATH}/image_cache_folder1\
+/// {CACHE_ROOT_PATH}/image_cache_folder2\
+/// etc.
+///
+/// An image cache folder contains:
+///
+/// {IMAGE_FOLDER_PATH}/config.json - the image configuration, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/manifest.json - the image manifest, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/layers - folder containing all layers, each in a separate gzip compressed tar file.
+#[derive(Clone)]
+pub struct CacheManager {
+    /// The root folder of the cache
+    root_path: PathBuf,
+
+    /// A map storing the cached images, with the map entry format being (image_reference, image_hash)
+    cached_images: HashMap<String, String>,
+}
+
+impl CacheManager {
+    /// Creates a new CacheManager instance and returns it. As argument, a path to the root folder
+    /// of the cache should be provided.
+    ///
+    /// Apart from that, the function also creates (if not already created) all folders from the path
+    /// specified as argument.
+    ///
+    /// If an index.json file exists at the path, it loads the file's contents into the 'cached_images'
+    /// field. If not, a new empty index.json file is created at that path.
+    pub fn new<P: AsRef<Path>>(root_path: P) -> Result<Self> {
+        // Create all missing folders, if not already created
+        fs::create_dir_all(&root_path).map_err(EnclaveBuildError::CacheInitError)?;
+
+        let mut contents = String::new();
+
+        // Try to open the index.json file and read the contents.
+        // If the file is missing, create it.
+        OpenOptions::new()
+            .write(true)
+            .read(true)
+            .create(true)
+            .open(root_path.as_ref().to_path_buf().join(CACHE_INDEX_FILE_NAME))
+            .map_err(EnclaveBuildError::CacheInitError)?
+            .read_to_string(&mut contents)
+            .map_err(EnclaveBuildError::CacheInitError)?;
+
+        // If the index.json file is empty, return an empty hashmap
+        if contents.is_empty() {
+            return Ok(Self {
+                root_path: root_path.as_ref().to_path_buf(),
+                cached_images: HashMap::new(),
+            });
+        }
+
+        // Try to deserialize the JSON string into a HashMap
+        let cached_images: HashMap<String, String> =
+            serde_json::from_str(&contents).map_err(EnclaveBuildError::SerdeError)?;
+
+        Ok(Self {
+            root_path: root_path.as_ref().to_path_buf(),
+            cached_images,
+        })
+    }
+
+    /// Stores the image data provided as argument in the cache at the folder pointed
+    /// by the 'root_path' field.
+    pub fn store_image_data<S: AsRef<str>>(
+        &mut self,
+        image_name: S,
+        image_data: &ImageData,
+    ) -> Result<()> {
+        let image_hash = image::image_hash(image_data.config.data.as_slice()).map_err(|err| {
+            EnclaveBuildError::HashingError(format!("Config hashing failed: {:?}", err))
+        })?;
+
+        // Create the folder where the image data will be stored
+        let target_path = self.root_path.join(&image_hash);
+        fs::create_dir_all(&target_path).map_err(EnclaveBuildError::CacheStoreError)?;
+
+        // Create the 'layers' folder and store the layers in it
+        let layers_path = target_path.join(CACHE_LAYERS_FOLDER_NAME);
+        fs::create_dir_all(&layers_path).map_err(EnclaveBuildError::CacheStoreError)?;
+
+        for layer in &image_data.layers {
+            // Each layer file will be named after the layer's digest hash
+            let layer_file_path =
+                layers_path.join(format!("{:x}", sha2::Sha256::digest(&layer.data)));
+            File::create(&layer_file_path)
+                .map_err(EnclaveBuildError::CacheStoreError)?
+                .write_all(&layer.data)
+                .map_err(EnclaveBuildError::CacheStoreError)?;
+        }
+
+        // Store the manifest
+        let manifest_json = match &image_data.manifest {
+            Some(image_manifest) => {
+                serde_json::to_string(&image_manifest).map_err(EnclaveBuildError::SerdeError)
+            }
+            None => Err(EnclaveBuildError::ManifestError),
+        }?;
+
+        File::create(&target_path.join(CACHE_MANIFEST_FILE_NAME))
+            .map_err(EnclaveBuildError::CacheStoreError)?
+            .write_all(manifest_json.as_bytes())
+            .map_err(EnclaveBuildError::CacheStoreError)?;
+
+        // Store the config
+        let config_json = String::from_utf8(image_data.config.data.clone()).map_err(|_| {
+            EnclaveBuildError::CacheStoreError(Error::new(
+                ErrorKind::InvalidData,
+                "Config data invalid",
+            ))
+        })?;
+
+        File::create(&target_path.join(CACHE_CONFIG_FILE_NAME))
+            .map_err(EnclaveBuildError::CacheStoreError)?
+            .write_all(config_json.as_bytes())
+            .map_err(EnclaveBuildError::CacheStoreError)?;
+
+        // If all image data was successfully stored, add the image to the index.json file and
+        // the hashmap
+        let image_ref = image::build_image_reference(image_name)?;
+        self.cached_images.insert(image_ref.whole(), image_hash);
+
+        let index_file = File::options()
+            .write(true)
+            .open(self.root_path.join(CACHE_INDEX_FILE_NAME))
+            .map_err(EnclaveBuildError::CacheStoreError)?;
+
+        // Write the hashmap (the image URI <-> image hash mappings) to the index.json file
+        serde_json::to_writer(index_file, &self.cached_images)
+            .map_err(EnclaveBuildError::SerdeError)?;
+
+        Ok(())
+    }
+
+    /// Determines if an image is stored correctly in the cache represented by the current CacheManager object.
+    pub fn check_cached_image<S: AsRef<str>>(&self, image_name: S) -> Result<()> {
+        // Check that the index.json file exists
+        let index_file_path = self.root_path.join(CACHE_INDEX_FILE_NAME);
+        fs::metadata(&index_file_path).map_err(|err| {
+            EnclaveBuildError::CacheMissError(format!("Cache index file missing: {:?}", err))
+        })?;
+
+        // If the image is not in the index.json file, then it is definitely not cached
+        let image_hash = self.get_image_hash_from_name(&image_name).ok_or_else(|| {
+            EnclaveBuildError::CacheMissError(
+                "Image hash missing from index.json file.".to_string(),
+            )
+        })?;
+
+        // The image is theoretically cached, but check the manifest, config and layers to validate
+        // that the image data is stored correctly
+        let image_folder_path = self.root_path.join(&image_hash);
+
+        // First validate the manifest
+        // Since the struct pulled by the oci_distribution API does not contain the manifest digest,
+        // and another HTTP request should be made to get the digest, just check that the manifest file
+        // exists and is not empty
+        let manifest_str = self
+            .fetch_manifest(&image_name)
+            .map_err(|_| EnclaveBuildError::ManifestError)?;
+
+        // The manifest is checked, so now validate the layers
+        self.validate_layers(
+            &image_folder_path.join(CACHE_LAYERS_FOLDER_NAME),
+            &manifest_str,
+        )?;
+
+        // Finally, check that the config is correctly cached
+        // This is done by applying a hash function on the config file contents and comparing the
+        // result with the config digest from the manifest
+        let config_str = self.fetch_config(&image_name)?;
+
+        let manifest_obj: serde_json::Value =
+            serde_json::from_str(manifest_str.as_str()).map_err(|_| {
+                EnclaveBuildError::CacheMissError("Could not parse manifest JSON.".to_string())
+            })?;
+
+        // Extract the config digest hash from the manifest
+        let config_digest = manifest_obj
+            .get("config")
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "'config' field missing from image manifest.".to_string(),
+                )
+            })?
+            .get("digest")
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "'digest' field missing from image manifest.".to_string(),
+                )
+            })?
+            .as_str()
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "Failed to get config digest from image manifest.".to_string(),
+                )
+            })?
+            .strip_prefix("sha256:")
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "Failed to get config digest from image manifest.".to_string(),
+                )
+            })?
+            .to_string();
+        // Compare the two digests
+        if config_digest != format!("{:x}", sha2::Sha256::digest(config_str.as_bytes())) {
+            return Err(EnclaveBuildError::CacheMissError(
+                "Config content digest and manifest digest do not match".to_string(),
+            ));
+        }
+
+        Ok(())
+    }
+
+    /// Fetches the image metadata from cache as an ImageDetails struct.
+    ///
+    /// If the data is not correctly cached or a file is missing, it returns an error.
+    ///
+    /// If the image is not cached, it does not attempt to pull the image from remote.
+    pub fn fetch_image_details<S: AsRef<str>>(&self, image_name: S) -> Result<ImageDetails> {
+        let hash = self
+            .get_image_hash_from_name(&image_name)
+            .ok_or_else(|| EnclaveBuildError::CacheMissError("Image hash missing".to_string()))?;
+
+        // Add algorithm prefix to the hash
+        let image_hash = format!("sha256:{}", hash);
+
+        let config_json = self.fetch_config(&image_name)?;
+
+        Ok(ImageDetails::new(
+            image::build_image_reference(&image_name)?.whole(),
+            image_hash,
+            image::deserialize_from_reader(config_json.as_bytes())?,
+        ))
+    }
+
+    /// Returns the manifest JSON string from the cache.
+    pub fn fetch_manifest<S: AsRef<str>>(&self, image_name: S) -> Result<String> {
+        let target_path = self.get_image_folder_path(&image_name)?;
+
+        // Read the JSON string from the cached manifest file
+        let manifest_path = target_path.join(CACHE_MANIFEST_FILE_NAME);
+        let mut manifest_json = String::new();
+        File::open(&manifest_path)
+            .map_err(|_| EnclaveBuildError::ManifestError)?
+            .read_to_string(&mut manifest_json)
+            .map_err(|_| EnclaveBuildError::ManifestError)?;
+
+        if manifest_json.is_empty() {
+            return Err(EnclaveBuildError::ManifestError);
+        }
+
+        Ok(manifest_json)
+    }
+
+    /// Returns the config JSON string from the cache.
+    pub fn fetch_config<S: AsRef<str>>(&self, image_name: S) -> Result<String> {
+        let target_path = self.get_image_folder_path(&image_name)?;
+
+        let mut config_json = String::new();
+        File::open(target_path.join(CACHE_CONFIG_FILE_NAME))
+            .map_err(|_| EnclaveBuildError::ConfigError)?
+            .read_to_string(&mut config_json)
+            .map_err(|_| EnclaveBuildError::ConfigError)?;
+
+        if config_json.is_empty() {
+            return Err(EnclaveBuildError::ConfigError);
+        }
+
+        Ok(config_json)
+    }
+
+    /// Validates that the image layers are cached correctly by checking them with the layer descriptors
+    /// from the image manifest.
+    fn validate_layers<P: AsRef<Path>>(&self, layers_path: P, manifest_str: &str) -> Result<()> {
+        let manifest_obj: serde_json::Value = serde_json::from_str(manifest_str).map_err(|_| {
+            EnclaveBuildError::CacheMissError("Manifest serialization failed".to_string())
+        })?;
+
+        // Try to get the layer list from the manifest JSON
+        let layers_vec = manifest_obj
+            .get("layers")
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "'layers' field missing from manifest JSON.".to_string(),
+                )
+            })?
+            .as_array()
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError("Manifest deserialize error.".to_string())
+            })?
+            .to_vec();
+
+        // Get the cached layers as a HashMap mapping a layer digest to the corresponding layer file
+        let mut cached_layers = HashMap::new();
+
+        fs::read_dir(layers_path)
+            .map_err(|err| {
+                EnclaveBuildError::CacheMissError(format!("Failed to get image layers: {:?}", err))
+            })?
+            .into_iter()
+            // Get only the valid directory entries that are valid files and return (name, file) pair
+            .filter_map(|entry| match entry {
+                Ok(dir_entry) => match File::open(dir_entry.path()) {
+                    Ok(file) => Some((dir_entry.file_name(), file)),
+                    Err(_) => None,
+                },
+                Err(_) => None,
+            })
+            // Map a layer digest to the layer file
+            // The 'cached_layers' hashmap will contain all layer files found in the cache for the current image
+            .for_each(|(name, file)| {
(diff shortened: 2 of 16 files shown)


enclave_build/src/utils.rs
```diff
@@ -0,0 +1,431 @@
...
+            digest: None,
+            config: config_obj,
+            manifest: Some(manifest_obj)
+        };
+
+        Ok(image_data)
+    }
+
+    #[test]
+    fn test_extract_layers() -> Result<(), ExtractTestError> {
```
sabin-rapan, line 332: tests assert, they dont return results/errors

sabin-rapan (changes requested):

enclave_build/src/utils.rs
```diff
@@ -0,0 +1,431 @@
...
+
+        Ok(res)
+    }
+}
+
+
+// TESTING
+
+#[derive(Debug, PartialEq)]
+pub enum ExtractTestError {
```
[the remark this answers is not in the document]
petreeftime: If it's only used in tests, then it should be defined in `mod tests`.

enclave_build/src/utils.rs
```diff
@@ -0,0 +1,431 @@
...
+    EnvCmdError(String),
+    ImageHashError(String),
+}
+
+/// Contains the logic for the extraction of image data from an ImageData struct to be stored later
+/// in the local cache
+/// 
+/// The ImageData struct represents the data of an image and is what the pull operation from
+/// https://github.com/krustlet/oci-distribution returns
+pub struct ExtractLogic {}
```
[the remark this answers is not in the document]
petreeftime: Indeed. Perhaps `ImageExtractor` would be a better name. And after looking at the code, it should contain ImageData or a reference to ImageData.

enclave_build/src/utils.rs
```diff
@@ -0,0 +1,431 @@
...
+
+impl ExtractLogic {
```
petreeftime, line 29: This looks like the code in extract.rs, just wrapped wrapped in an impl? Is there something I'm missing here? Why duplicate this code?

petreeftime (changes requested):

enclave_build/src/extract.rs
```diff
@@ -0,0 +1,415 @@
...
+                Ok(cmd_strings)
+            }
+    }
+}
+
+/// Extract the image hash (digest) from an image
+pub fn extract_image_hash(image_data: &ImageData) -> Result<String> {
+    // Extract the config JSON from the image
+    let config_json = extract_config_json(image_data)
+        .map_err(|err| ExtractError::ImageHashError(format!("{:?}", err)))?;
```
sabin-rapan, line 153: This could have easily been:
```
 .map_err(|err| ExtractError::ImageHashError(err))?
```

Same for ConfigError(String) EnvCmdError(String)

sabin-rapan (changes requested):

enclave_build/Cargo.toml
```diff
@@ -21,4 +21,5 @@ url = "2.1"
 sha2 = "0.9.5"
 futures = "0.3.13"
 
+oci-distribution = "0.9.2"
```
sabin-rapan, line 24: CI build fails with
```
 Compiling oci-distribution v0.9.2
error[E0658]: destructuring assignments are unstable
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/oci-distribution-0.9.2/src/client.rs:379:31
    |
379 |             (location, start) = self.push_chunk(&location, image, blob_data, start).await?;
    |             ----------------- ^
    |             |
    |             cannot assign to this expression
    |
    = note: see issue #71126 <https://github.com/rust-lang/rust/issues/71126> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `oci-distribution` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
make: *** [.build-nitro-cli] Error 101 Compiling oci-distribution v0.9.2
error[E0658]: destructuring assignments are unstable
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/oci-distribution-0.9.2/src/client.rs:379:31
    |
379 |             (location, start) = self.push_chunk(&location, image, blob_data, start).await?;
    |             ----------------- ^
    |             |
    |             cannot assign to this expression
    |
    = note: see issue #71126 <https://github.com/rust-lang/rust/issues/71126> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `oci-distribution` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
make: *** [.build-nitro-cli] Error 101
```

it looks like they built the crate with nightly.
we build with stable.
maybe try to find an older version that works with stable

enclave_build/src/constants.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2019-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use oci_distribution::manifest;
+
+/// Docker inspect architecture constants
+pub const DOCKER_ARCH_ARM64: &str = "arm64";
```
shtaked, line 7: are those constants in use somewhere?

calincoman: No, they were initially in docker.rs and I moved them here, will remove them in next commit.

enclave_build/src/image.rs
```diff
@@ -0,0 +1,287 @@
...
+    Clone, Debug,
+    Default,
+    Serialize,
+    Deserialize,
+    Eq,
+    PartialEq,
+    Getters,
+    Setters,
+)]
+ #[getset(get = "pub", set = "pub")]
```
sabin-rapan, line 212: If we had code coverage (which we don't, but still), it would be impossible to get a decent percentage with all these setters/getters.
You would have to write unit tests that set/get each field. That's annoying and waste of code.

My recommendation would be to write setters/getters manually only for the things you actually need, or if you need all of the fields, then make them public (not the best coding pattern, but it's the lesser of two evils).

enclave_build/src/image.rs
```diff
@@ -0,0 +1,287 @@
...
+    diff_ids: Vec<String>,
+}
+
+#[derive(
+    Clone, Debug,
+    Default,
+    Serialize,
+    Deserialize,
+    Eq,
+    PartialEq,
```
sabin-rapan, line 208: Except Serialize, Deserialize do we actually need all these default implementations for all the structs?

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,92 @@
+// Copyright 2019-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sabin-rapan, line 1: We are in 2022.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+fn serialize_to_file<P: AsRef<Path>, T: Serialize>(path: P, object: &T, pretty: bool) -> CacheResult<()> {
+    let path = path.as_ref();
+    let file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(path)
+            .map_err(|err| CacheError::ReadWriteError(err))?;
```
sabin-rapan, line 58: Is this how cargo fmt aligned the code?

enclave_build/src/image.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    os: Option<String>,
+
+    #[serde(rename = "os.version")]
+    os_version: Option<String>,
+
+    #[serde(rename = "os.features")]
+    os_features: Option<Vec<String>>,
+
+    config: Option<Config>,
```
sabin-rapan, line 120: Do we want this to be serialized as "Option::None" ?

enclave_build/Cargo.toml
```diff
@@ -14,11 +14,14 @@ serde_yaml = "0.8"
...
-tokio = { version = "1.0", features = ["rt-multi-thread"] }
 base64 = "0.13"
 log = "0.4"
 url = "2.1"
 sha2 = "0.9.5"
 futures = "0.3.13"
 
 aws-nitro-enclaves-image-format = "0.1"
+
+getset = "0.1.2"
```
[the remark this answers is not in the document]
lauralt: Ah, tokio was already used in the project.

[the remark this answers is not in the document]
calincoman: Yes, and I am using it in testing async functions too (not yet committed).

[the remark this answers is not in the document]
lauralt: What about getset? Can we remove this from the deps list?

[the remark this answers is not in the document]
calincoman: Yes, will remove it, I was using it previously to generate getters and setters.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,208 @@
...
+                Error::CredentialsError(format!("Invalid Base64 encoding for auth: {}", err))
+            })?;
+            let decoded = std::str::from_utf8(&decoded).map_err(|err| {
+                Error::CredentialsError(format!("Invalid utf8 encoding for auth: {}", err))
+            })?;
+
+            // Try to get the username and the password
+            if let Some(index) = decoded.rfind(':') {
+                let (username, after_user) = decoded.split_at(index);
+                let (_, password) = after_user.split_at(1);
```
lauralt, line 136: magic value

lauralt (commented):
Left a few Rust style related comments. I haven't looked over the tests yet.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,121 @@
...
+fn deserialize_from_file<P: AsRef<Path>, T: DeserializeOwned>(path: P) -> Result<T> {
+    let path = path.as_ref();
+    let manifest_file =
+        std::io::BufReader::new(fs::File::open(path).map_err(|err| Error::ReadWriteError(err))?);
+    let manifest = serde_json::from_reader(manifest_file).map_err(|err| Error::SerdeError(err))?;
+
+    Ok(manifest)
+}
+
+fn deserialize_from_reader<R: Read, T: DeserializeOwned>(reader: R) -> Result<T> {
```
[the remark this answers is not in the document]
calincoman: Yes, there is a possibility I might use them.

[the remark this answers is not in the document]
lauralt: It's a good practice to add things when you are also using them, or at least add a disclaimer or something: `// These will be used in tests, and moved in the tests module.`

sabin-rapan (commented):
You should take some time to make sure the CI passes on your PR.

calincoman:
> You should take some time to make sure the CI passes on your PR.

Yes, I am looking on the tests now.

enclave_build/src/image.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    Ok(hash)
+}
+
+/// Builds a docker image reference from the image name given as parameter.
+///
+/// e.g. "hello-world" image has reference "docker.io/library/hello-world:latest".
+///
+/// Uses the implementation from oci_distribution.
+pub fn image_reference<S: AsRef<str>>(image_name: S) -> Result<Reference, Error> {
```
sabin-rapan, line 124: You really like to type Error every time instead of declaring your own Result type that wraps standard Result and your Error enum

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub async fn get_image_details<S: AsRef<str>>(
+        &mut self,
+        image_name: S,
+    ) -> Result<ImageDetails> {
+        let local_cache = (&mut self.cache).as_mut();
+
+        // If the image is cached, fetch and return its metadata
+        if local_cache.is_some() && local_cache.as_ref().unwrap().is_cached(&image_name).is_ok() {
+            let image_details = local_cache.unwrap().fetch_image_details(&image_name)?;
+            println!("Image found in local cache.");
```
sabin-rapan, line 96: debugging leftover?

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,784 @@
...
+                    home_folder
+                        .into_string()
+                        .map_err(|err| Error::CacheRootPath(format!("{:?}", err)))?
+                )
+            }
+        };
+
+        let mut path = PathBuf::from(root);
+        // Add the additional path to the root
+        path.push(".nitro_cli/container_cache");
```
petreeftime, line 495: use `aws-nitro-enclaves-cli` instead of `.nitro_cli`. There's already another `nitro-cli` package in Ubuntu/Fedora and they shouldn't collide.

enclave_build/src/lib.rs
```diff
@@ -58,8 +63,15 @@ pub enum Docker2EifError {
...
 impl<'a> Docker2Eif<'a> {
+    /// If the 'docker_dir' argument is Some (i.e. --docker-dir flag is used), the docker daemon will
+    /// always be used to build the image locally and store it in the docker cache.
+    ///
+    /// If the 'oci_image' argument is Some (i.e. the --image-uri flag is used) and the '--docker-dir'
+    /// flag is not used, pull and cache the image without using the docker daemon.
     pub fn new(
         docker_image: String,
+        docker_dir: Option<String>,
+        oci_image: Option<String>,
```
petreeftime, line 74: Shouldn't these 3 be exclusive to each other? Perhaps use a single enum instead?

enclave_build/src/image.rs
```diff
@@ -0,0 +1,136 @@
...
+        })
+    }
+}
+
+/// Calculates the image ID (or image hash) as the SHA256 digest of the image config.
+///
+/// This method is described in the OCI image spec.
+///
+/// https://github.com/opencontainers/image-spec/blob/main/config.md
+pub fn image_hash<R: Read>(mut image_config: R) -> Result<String, Error> {
```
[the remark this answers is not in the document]
calincoman: I am using it in cache.rs
https://github.com/calincoman/aws-nitro-enclaves-cli/blob/feature_fork_branch/enclave_build/src/cache.rs#L172

enclave_build/src/lib.rs
calincoman: For backwards compatibility I kept the '--docker-uri' flag required, and if the '--oci-uri' flag is used, then the argument of the first one is ignored.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,784 @@
...
+pub const CACHE_INDEX_FILE_NAME: &str = "index.json";
+/// The name of the folder used by the cache to store the image layers
+pub const CACHE_LAYERS_FOLDER_NAME: &str = "layers";
+/// The name of the image config file from the cache
+pub const CACHE_CONFIG_FILE_NAME: &str = "config.json";
+/// The name of the image manifest file from the cache
+pub const CACHE_MANIFEST_FILE_NAME: &str = "manifest.json";
+
+#[derive(Debug)]
+pub enum Error {
```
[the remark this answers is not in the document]
calincoman: I tried to make the errors as specific as possible.

enclave_build/src/lib.rs
petreeftime: That is quite confusing IMO. The CLI shouldn't accept both --docker-uri and --oci-uri at the same time.

petreeftime: It should be either --docker-uri with optional --docker-dir or --oci-uri

enclave_build/src/cache.rs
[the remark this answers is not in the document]
petreeftime: They can be specific, but right now there is some overlap between all the error enums. Generally crates expose a single Error-type / Result-type, so that someone who calls on the crate can implement a more unified way of dealing with them.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,196 @@
...
+            let mut _user = String::new();
+            let mut _pass = String::new();
+            println!("Credentials found.");
+            if let RegistryAuth::Basic(user, pass) = registry_auth {
+                println!("Username: {}", user);
+                println!("Password: {}", pass);
+                _user = user;
+                _pass = pass;
+            }
+            RegistryAuth::Basic(_user, _pass)
```
lauralt, line 163: the `_`s are pretty unusual. You can replace this block with:
```rust
            println!("Credentials found.");
            let (user, pass) = if let RegistryAuth::Basic(user, pass) = registry_auth {
                println!("Username: {}", user);
                println!("Password: {}", pass);
                (user, pass)
            } else {
                (String::new(), String::new())
            };
            RegistryAuth::Basic(user, pass)
```

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,196 @@
...
+/// Pulls an image (all blobs - layers, manifest and config) from the Docker remote registry.
+///
+/// The function takes as argument the image name (e.g. hello-world, postgres, ubuntu etc.)
+pub async fn pull_image_data<S: AsRef<str>>(image_name: S) -> Result<ImageData> {
+    // Build the client required for the pulling - uses HTTPS protocol
+    let mut client = build_client(ClientProtocol::Https);
+
+    // Build an image reference from the image name
+    let image_ref =
+        image::image_reference(image_name).map_err(|err| Error::Image(format!("{:?}", err)))?;
```
lauralt, line 182: I find `err.to_string()` less verbose than the `format!` solution.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    // Pull an ImageData struct containing the layers, manifest and configuration file
+    let image_data = client
+        .pull(&image_ref, &auth, ACCEPTED_MEDIA_TYPES.to_vec())
+        .await;
+
+    match image_data {
+        Ok(img_data) => Ok(img_data),
+        Err(err) => Err(Error::Image(err.to_string())),
+    }
```
lauralt, line 195:
```rust
    client
        .pull(&image_ref, &auth, ACCEPTED_MEDIA_TYPES.to_vec())
        .await
        .map_err(|err| Error::Image(err.to_string()))
```

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,196 @@
...
+            RegistryAuth::Anonymous
+        }
+    }
+}
+
+/// Pulls an image (all blobs - layers, manifest and config) from the Docker remote registry.
+///
+/// The function takes as argument the image name (e.g. hello-world, postgres, ubuntu etc.)
+pub async fn pull_image_data<S: AsRef<str>>(image_name: S) -> Result<ImageData> {
+    // Build the client required for the pulling - uses HTTPS protocol
```
lauralt, line 177: The comments in this function are mostly superfluous.

lauralt (commented):
Reviewed the first two commits for now.

enclave_build/src/pull.rs
petreeftime: Also, don't use `println!` for outputing stuff. Use the `log` interface instead. And definitely do not log a `password`.

calincoman: Yes, that was a debugging leftover, sorry for that.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,782 @@
...
+                Error::ImageNotCached("'layers' field missing from manifest JSON.".to_string())
+            })?
+            .as_array()
+            .ok_or_else(|| Error::ImageNotCached("Manifest deserialize error.".to_string()))?
+            .to_vec();
+
+        // Get the cached layers as a HashMap mapping a layer digest to the corresponding layer file
+        let mut cached_layers: HashMap<String, File> = HashMap::new();
+
+        fs::read_dir(layers_path)
```
lauralt, line 413: Wow 🥺 I need a fresh mind to review this, but I see some undocumented unwraps :(.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,782 @@
...
+        let image_hash = self.get_image_hash_from_name(&image_name);
+        if image_hash.is_none() {
+            return Err(Error::ImageNotCached(
+                "Image hash missing from index.json file.".to_string(),
+            ));
+        }
+        // The image is theoretically cached, but check the manifest, config and layers to validate
+        // that the image data is stored correctly
+
+        let image_folder_path = self.root_path.clone().join(&image_hash.unwrap());
```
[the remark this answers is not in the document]
calincoman: In the if statement on line 263, I exit the function with an error if the image_hash is None
Screenshot 2022-09-27 at 19 19 32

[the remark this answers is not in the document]
lauralt: Yeah, I meant explain in the code, all the unwraps should be documented even if it's obvious they are safe.

enclave_build/src/cache.rs
calincoman: Before each unwrap i am doing a .filter in order to keep only the Ok() variants

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,312 @@
...
+pub struct OciImageManager {
+    /// Name of the container image.
+    image_name: String,
+    /// Have the cache as an option in order to not stop the CLI if there is a cache creation
+    /// error (the image will simply be pulled in this case, instead of being fetched from cache)
+    cache: Option<CacheManager>,
+}
+
+impl ImageManager for OciImageManager {
+    fn get_image_name(&self) -> &String {
```
lauralt, line 87: Getters in Rust usually don't include the `get_` part in their name: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter.

enclave_build/src/lib.rs
```diff
@@ -58,8 +63,15 @@ pub enum Docker2EifError {
 }
 
 impl<'a> Docker2Eif<'a> {
+    /// If the 'docker_dir' argument is Some (i.e. --docker-dir flag is used), the docker daemon will
+    /// always be used to build the image locally and store it in the docker cache.
+    ///
+    /// If the 'oci_image' argument is Some (i.e. the --image-uri flag is used) and the '--docker-dir'
+    /// flag is not used, pull and cache the image without using the docker daemon.
     pub fn new(
```
lauralt, line 71: Wow, this function has a lot of parameters. It is a good practice to limit the number of them to 3-4. Isn't clippy complaining about this?

calincoman: Clippy is not complaining, but this is the original function, I added just 2 new parameters (docker_dir and oci_image).

andreeaflorescu: Clippy is actually complaining, but it was silenced by writing allow on line 3 of this file. Since your parameters are optional anyway you can avoid introducing them in the constructor and use a more idiomatic way by initializing them to None when creating the structure and have separate setters for them that can be called only if they're need it.

BTW, directories and paths should not have the type `String`, but should use `Path` or `PathBuf` instead. Paths do not follow the same requirements as strings do.

enclave_build/src/lib.rs
```diff
@@ -12,14 +12,63 @@ mod yaml_generator;
...
+    #[error("Config error")]
+    ConfigError,
+    #[error("Manifest error")]
+    ManifestError,
+    #[error("Failed to extract expressions from image: `{0}`")]
+    ExtractError(String),
+    #[error("EIF build error: `{0}`")]
+    OtherError(String),
+}
+
```
raulmldv, line 79: Removed specific errors from each file. Centralized the error types into `EnclaveBuildError` without the need to implement the `Display` trait for each of them as we use `thiserror`. Modified each occurrence of these errors.

enclave_build/src/lib.rs
raulmldv: I created structures for grouping the parameters before passing them. There are too few optional fields to make for a builder pattern.
For now I left the paths as Strings as they are represented like this throughout the whole repo. It might be the focus of another PR to propagate `PathBufs`.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,193 @@
...
+        ..Default::default()
+    };
+    Client::new(client_config)
+}
+
+/// Checks the DOCKER_CONFIG env variable for the path to the docker config file
+///
+/// If DOCKER_CONFIG is not set, try to get the config file from the default path
+/// {XDG_RUNTIME_DIR}/containers/auth.json or {HOME}/.docker/config.json in a Podman-like behavior
+fn get_docker_config_file() -> Result<File> {
```
raulmldv, line 43: Added Podman-like behaviour for searching the credentials. After checking the `DOCKER_COFIG` env variable, the search falls back in order to `{XDG_RUNTIME_DIR}/containers/auth.json` or `{HOME}/.docker/config.json`.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,196 @@
...
+///
+/// The assumed format of the file is:\
+/// {\
+///        "auths": {\
+///            "https://index.docker.io/v1/": {\
+///                    "auth": "<token_string>"\
+///            }\
+///        }\
+/// }
+pub fn parse_credentials() -> Result<RegistryAuth> {
```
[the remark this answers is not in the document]
raulmldv: Modified `get_docker_config_file` function.

enclave_build/src/cache.rs
raulmldv: Simplified map creation with a `filer_map` and got rid of unwraps.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Returns a struct containing image metadata.
+    ///
+    /// If the image is cached correctly, the function tries to fetch the image from the cache.
+    ///
+    /// If the image is not cached or a cache was not created (the 'cache' field is None),
+    /// it pulls the image, caches it (if the 'cache' field is not None) and returns its metadata.
+    ///
+    /// If the pull succeeded but the caching failed, it returns the pulled image metadata.
+    async fn get_image_details<S: AsRef<str>>(&mut self, image_name: S) -> Result<ImageDetails> {
```
raulmldv, line 149: Modified workflow to avoid unwraps and treat the following case: if we have a missing or invalid cache entry, pull image and store it in the cache.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,247 @@
...
+            .ok_or(EnclaveBuildError::ConfigError)?
+            .env();
+        let entrypoint = image
+            .config
+            .config()
+            .as_ref()
+            .ok_or(EnclaveBuildError::ConfigError)?
+            .entrypoint();
+
+        // If no CMD instructions are found, try to locate an ENTRYPOINT command
```
raulmldv, line 235: Simplified if cascade.

enclave_build/src/lib.rs
```diff
@@ -107,7 +121,25 @@ impl<'a> Docker2Eif<'a> {
         metadata_path: Option<String>,
         build_info: EifBuildInfo,
     ) -> Result<Self> {
-        let docker = DockerUtil::new(docker_image.clone());
+        // The flags usage was already validated by the commands parser, so now just check if the docker daemon
+        // should be used or not
```
raulmldv, line 140: Properly match docker(less) case.

enclave_build/src/lib.rs
```diff
@@ -80,6 +79,33 @@ pub enum EnclaveBuildError {
...
+    pub img_version: Option<String>,
+    pub metadata_path: Option<String>,
+    pub build_info: EifBuildInfo,
+}
+
+pub struct SourceArgs {
+    pub docker_image: Option<String>,
+    pub docker_dir: Option<String>,
+    pub oci_image: Option<String>,
+}
```
raulmldv, line 107: Created substructures for building the `Docker2Eif` structure.

raulmldv (commented):
PR clean up

enclave_build/src/lib.rs
lauralt: When there are changes that are not necessarily related to the scope of the PR (and in this case also made by another person) such as this refactoring it's nice to add a separate commit for that.

enclave_build/src/image.rs
```diff
@@ -0,0 +1,131 @@
...
+        })
+    }
+}
+
+/// Calculates the image ID (or image hash) as the SHA256 digest of the image config.
+///
+/// This method is described in the OCI image spec.
+///
+/// https://github.com/opencontainers/image-spec/blob/main/config.md
+pub fn image_hash<R: Read>(mut image_config: R) -> Result<String> {
```
lauralt, line 64: This method seems to be a bit useless. We are using it only with slice parameters, for which we can just do the `format!` below, so no need to have a new Vec and move data around.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,734 @@
...
+    ///
+    /// If an index.json file exists at the path, it loads the file's contents into the 'cached_images'
+    /// field. If not, a new empty index.json file is created at that path.
+    pub fn new<P: AsRef<Path>>(root_path: P) -> Result<Self> {
+        // Create all missing folders, if not already created
+        fs::create_dir_all(&root_path).map_err(EnclaveBuildError::CacheInitError)?;
+
+        let mut contents = String::new();
+
+        // Try to open the index.json file and read the contents.
```
lauralt, line 71: I find it pretty confusing to define a constant for some string, and then in comments use that particular string. We could instead say "try to open the cache index file".

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,734 @@
...
+        if contents.is_empty() {
+            return Ok(Self {
+                root_path: root_path.as_ref().to_path_buf(),
+                cached_images: HashMap::new(),
+            });
+        }
+
+        // Try to deserialize the JSON string into a HashMap
+        let cached_images: HashMap<String, String> =
+            serde_json::from_str(&contents).map_err(EnclaveBuildError::SerdeError)?;
```
lauralt, line 92: You can replace this with:

```rust
        let cached_images = if contents.is_empty() {
            HashMap::new()
        } else {
            serde_json::from_str(&contents).map_err(EnclaveBuildError::SerdeError)?
        };
```

enclave_build/src/image.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    /// Try to build an ImageDetails struct from an oci_distribution ImageData struct.
+    //
+    // The oci_distribution ImageData struct does not contain the image name or reference, so this
+    // must be additionally passed to the function as well.
+    pub fn from<S: AsRef<str>>(image_name: S, image_data: &ImageData) -> Result<Self> {
+        // Get the config JSON String from the pulled image
+        let config_json = String::from_utf8(image_data.config.data.clone())
```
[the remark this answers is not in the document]
lauralt: Ah so I assume we want to check if the bytes are valid utf-8 here. We should add a comment about that if that's the case.

enclave_build/src/image.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub uri: String,
+    /// The image ID, calculated as the SHA256 digest hash of the image config.
+    #[serde(rename = "Id")]
+    pub hash: String,
+    /// The image config.
+    pub config: ImageConfiguration,
+}
+
+impl ImageDetails {
+    pub fn new<S: AsRef<str>>(image_uri: S, image_hash: S, config: ImageConfiguration) -> Self {
```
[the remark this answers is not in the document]
raulmldv: Seems like the only reason is to allow using `String` and `&str` across tests, will remove it to simplify.

enclave_build/src/image.rs
```diff
@@ -0,0 +1,131 @@
...
+        Self {
+            uri: image_uri.as_ref().to_string(),
+            hash: image_hash.as_ref().to_string(),
+            config,
+        }
+    }
+
+    /// Try to build an ImageDetails struct from an oci_distribution ImageData struct.
```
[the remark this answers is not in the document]
raulmldv: As `from` just uses the config field from the details, I can remove the function altogether and use `new` instead.

enclave_build/src/image.rs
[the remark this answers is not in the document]
raulmldv: Don't think it's needed just for utf8 validation, we can convert directly.

enclave_build/src/lib.rs
raulmldv: Will make separate initial PR.

enclave_build/src/lib.rs
```diff
@@ -250,24 +286,29 @@ impl<'a> Docker2Eif<'a> {
...
             eprintln!("Docker error: {:?}", e);
-            Docker2EifError::DockerError
+            EnclaveBuildError::DockerError(e)
         })?;
 
         let flags = match arch.as_str() {
             docker::DOCKER_ARCH_ARM64 => EIF_HDR_ARCH_ARM64,
             docker::DOCKER_ARCH_AMD64 => 0,
-            _ => return Err(Docker2EifError::UnsupportedArchError),
+            _ => {
```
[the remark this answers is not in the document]
raulmldv: Nope, just part of the error type change.

raulmldv:
Split this PR in 5 smaller ones for easier review and merging.

Issue #204, opened 2020-11-24
Signing enclave images with HSM/KSM


yoavj1:
Currently the only option to sign an enclave image is to pass the private key and the certificate to build-enclave command. However, it prevents storing the key in a secure storage like HSM or KSM, and using it only for signing without retrieving the key itself.
One solution is to allow to add the signature after the enclave was created, so the signature can be produced independently from the enclave creation.

petreeftime:
Thank you for the feature request. There are two topics here:
  1. Signing with an AWS KMS asymmetric key (and/or maybe a key from an HSM or Nitro Enclave :) ).
  2. Attaching/replacing a signature to an already created image.

gal-fordefi:
@petreeftime I guess that now when you merged https://github.com/awslabs/aws-nitro-enclaves-cose/issues/5 it's possible to add the requested functionality to the CLI.

petreeftime:
Yes, it should be possible to add support for this.

PR #406, opened 2022-08-29
Implement command 'sign-eif'

pro-vlad:
*Issue #, if available:*
#204 
*Description of changes:*

- Implemented 'sign-eif' which can be used to sign an existing EIF by using a private key or KMS.
- Add new arguments for 'describe-eif' for extracting certificate information for images signed with KMS.
- Add new arguments for 'run-enclave' for validating KMS signed images.


src/common/commands_parser.rs
```diff
@@ -274,6 +280,137 @@ impl PcrArgs {
...
+    /// The region in which the KMS key resides.
+    pub region: Option<String>,
+    /// The KMS key id.
+    pub key_id: Option<String>,
+}
+
+impl SignArgs {
+    /// Construct a new `SignArg` instance from the given command-line arguments.
+    pub fn new_with(args: &ArgMatches) -> NitroCliResult<Self> {
+        let signing_method = parse_signing_method(args)
```
petreeftime, line 300: This `signing_method` argument is really not required. It should be indicated by either the presence of `key_id` or `private_key` and no need to have a separate argument for it.

src/common/commands_parser.rs
```diff
@@ -274,6 +280,137 @@ impl PcrArgs {
...
+            .map_err(|err| err.add_subaction("Parse signing method".to_string()))?;
+        let private_key = parse_private_key(args);
+        let region = parse_region(args);
+        let key_id = parse_key_id(args);
```
petreeftime, line 304: `key_id` is a bit hard to interpret as "KMS Key ID". Perhaps `kms_key_arn` make more sense. Also `region` is not very obvious.

src/lib.rs
```diff
@@ -659,6 +684,22 @@ macro_rules! create_app {
...
+                            .required(false)
+                            .conflicts_with("config"),
+                    )
+                    .arg(
+                        Arg::with_name("key-id")
+                            .long("key-id")
+                            .takes_value(true)
+                            .help("The KMS key id.")
+                            .required(false)
+                            .conflicts_with("config"),
```
petreeftime, line 694: Should also conflict with `private_key`.

src/lib.rs
```diff
@@ -659,6 +684,22 @@ macro_rules! create_app {
                             .required(false)
                             .conflicts_with("config"),
                     )
+                    .arg(
+                        Arg::with_name("region")
+                            .long("region")
+                            .takes_value(true)
+                            .help("The region in which the KMS key resides.")
+                            .required(false)
+                            .conflicts_with("config"),
```
petreeftime, line 686: Should conflict with `private_key`.

src/common/commands_parser.rs
petreeftime: `region` should also be optional. `kms_client` should load it from `.aws/config` or from `AWS_REGION` env variable. Perhaps there's some way to check if any of these are set or not.

gavinuhma:
@pro-vlad Thanks for this! Are you still planning to get it merged? We're looking for this feature

enclave_build/src/main.rs
```diff
@@ -133,9 +146,26 @@ fn main() {
...
         .map(|val| val.to_string());
-    let private_key = matches
-        .value_of("private_certificate")
+    let kms_key_arn = matches.value_of("kms-key-arn");
+    let kms_key_region = matches.value_of("kms-key-region");
+    let private_key_path = matches
+        .value_of("private_key")
         .map(|val| val.to_string());
+
+    let signing_key = match (kms_key_arn, private_key_path) {
```
dwmw2, line 155: This shouldn't just be a key path. It should accept PKCS#11 URIs and conform to david.woodhou.se/draft-woodhouse-cert-best-practice.html

PR #420, opened 2022-11-04
enclave_build: Error refactor

raulmldv:
Dockerless frature (1/5)

This is the first PR from splitting the initial one #404.

To avoid defining error types in each module file to be implemented for this feature, I created a general error type in the `lib.rs` file. In this way, the enclave build error can be used outside the crate and also extended easier.
By using `thiserror` attributes, we don't need display implementations for every defined error.

### DIFF
--- a/enclave_build/src/docker.rs
+++ b/enclave_build/src/docker.rs
@@ -1,15 +1,13 @@
 // Copyright 2019-2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::docker::DockerError::CredentialsError;
+use crate::{EnclaveBuildError, Result};
 use futures::stream::StreamExt;
-use log::{debug, error, info};
+use log::{debug, info};
 use serde_json::{json, Value};
 use shiplift::RegistryAuth;
 use shiplift::{BuildOptions, Docker, PullOptions};
-use std::fs::File;
-use std::io::Write;
-use std::path::Path;
+use std::{fs::File, io::Write, path::Path};
 use tempfile::NamedTempFile;
 use tokio::runtime::Runtime;
 use url::Url;
@@ -17,17 +15,6 @@ use url::Url;
 /// Docker inspect architecture constants
 pub const DOCKER_ARCH_ARM64: &str = "arm64";
 pub const DOCKER_ARCH_AMD64: &str = "amd64";
-
-#[derive(Debug, PartialEq, Eq)]
-pub enum DockerError {
-    BuildError,
-    InspectError,
-    PullError,
-    RuntimeError,
-    TempfileError,
-    CredentialsError(String),
-    UnsupportedEntryPoint,
-}
 
 /// Struct exposing the Docker functionalities to the EIF builder
 pub struct DockerUtil {
@@ -59,7 +46,7 @@ impl DockerUtil {
     /// we are parsing it correctly, so the parsing mechanism had been infered by
     /// reading a config.json created by:
     //         Docker version 19.03.2
-    fn get_credentials(&self) -> Result<RegistryAuth, DockerError> {
+    fn get_credentials(&self) -> Result<RegistryAuth> {
         let image = self.docker_image.clone();
         let host = if let Ok(uri) = Url::parse(&image) {
             uri.host().map(|s| s.to_string())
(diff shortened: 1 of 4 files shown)


shtaked (dismissed):

enclave_build/src/lib.rs
```diff
@@ -73,34 +110,37 @@ impl<'a> Docker2Eif<'a> {
         img_version: Option<String>,
         metadata_path: Option<String>,
         build_info: EifBuildInfo,
-    ) -> Result<Self, Docker2EifError> {
+    ) -> Result<Self> {
         let docker = DockerUtil::new(docker_image.clone());
 
         if !Path::new(&init_path).is_file() {
-            return Err(Docker2EifError::InitPathError);
+            return Err(EnclaveBuildError::PathError("init path".to_string()));
```
petreeftime, line 117: Would be useful to include the Path in the error message.

enclave_build/src/lib.rs
```diff
@@ -23,8 +23,8 @@ pub const DEFAULT_TAG: &str = "1.0";
 
 #[derive(Debug, Error)]
 pub enum EnclaveBuildError {
-    #[error("Docker error: `{0:?}`")]
-    DockerError(DockerError),
+    #[error("Docker error: `{0}`")]
+    DockerError(String),
```
petreeftime, line 27: Why not use `thiserror` for DockerError as well and transform it to a formatted string instead of having strings all over the place?

raulmldv: That would mean keeping the DockerError type in `docker.rs` and passing strings at that level. Wasn't the idea of the refactoring to get rid of specific error type definitions? I can define multiple docker errors here.

petreeftime: Sure, but this commit basically transforms many of the structured errors into unstructured ones via String.

enclave_build/src/docker.rs
```diff
@@ -197,14 +199,18 @@ impl DockerUtil {
 
                             if let Some(err_msg) = msg.get("error") {
                                 error!("{:?}", err_msg.clone());
-                                break Err(DockerError::PullError);
+                                break Err(EnclaveBuildError::DockerError(
+                                    "Failed to pull docker image".into(),
+                                ));
```
petreeftime, line 204: This would need to be `EnclaveBuildError::ImagePullError(err_msg)`

enclave_build/src/lib.rs
petreeftime: I gave an example above.

raulmldv: Right, I'll make it more structured.

enclave_build/src/lib.rs
```diff
@@ -12,14 +12,71 @@ mod yaml_generator;
...
+    #[error("Failed to convert image to EIF")]
+    ImageConvertError,
+    #[error("Hashing error: `{0}`")]
+    HashingError(String),
+    #[error("Serde error: `{0:?}`")]
+    SerdeError(serde_json::Error),
+    #[error("Error while parsing credentials: `{0}`")]
+    CredentialsError(String),
+    #[error("Image cache initialization error: `{0:?}`")]
+    CacheInitError(std::io::Error),
```
lauralt, line 59: There are some variants here that will be used in subsequent commits. Ideally we should add them at that point, but if that takes too much time, please mention in the commit message about this thing.

enclave_build/src/lib.rs
```diff
@@ -12,14 +12,71 @@ mod yaml_generator;
...
+    #[error("Ramfs error: `{0:?}`")]
+    RamfsError(YamlGeneratorError),
+    #[error("Signature error: `{0}`")]
+    SignError(String),
+    #[error("Metadata error: `{0}`")]
+    MetadataError(String),
+    #[error("Unsupported architecture")]
+    UnsupportedArchError,
+    #[error("Getting image detail failed: `{0}`")]
+    ImageDetailError(String),
```
lauralt, line 47: I wanted to leave this comment in the other PR as well but forgot about it. Why do we use `ImageDetail` instead of `ImageMetadata` in the structure name/error variant etc. `detail` makes me think of something that is not important, and I think I would prefer the latter option.

raulmldv: The structure name introduced in the next PR is `ImageDetails` to avoid confusion with the EIF metadata section. They are both metadata of some image type.

enclave_build/src/lib.rs
```diff
@@ -121,35 +163,27 @@ impl<'a> Docker2Eif<'a> {
...
+        self.docker.pull_image()?;
 
         Ok(())
     }
 
-    pub fn build_docker_image(&self, dockerfile_dir: String) -> Result<(), Docker2EifError> {
+    pub fn build_docker_image(&self, dockerfile_dir: String) -> Result<()> {
         if !Path::new(&dockerfile_dir).is_dir() {
-            return Err(Docker2EifError::DockerfilePathError);
+            return Err(EnclaveBuildError::PathError("Dockerfile path".to_string()));
```
shtaked, line 174:
```suggestion
            return Err(EnclaveBuildError::PathError(dockerfile_dir.to_string()));
```

shtaked (approved):

petreeftime (approved):

enclave_build/src/lib.rs
lauralt: Hmm okay, maybe we can find a better name though, but we can leave that for when we add the ImageDetails structure.

lauralt (approved):

andraprs (approved):

PR #421, opened 2022-11-09
enclave_build: OCI image operations

raulmldv:
Dockerless feature #404 - 2nd PR from splitting

To reproduce the Docker behaviour for OCI images, we have to implement the two main operations for building an enclave image:  pulling and inspecting. This PR implements the image representation using the `oci_spec` crate and the image operations using `oci_distribution`.

The two new modules, `image` and `pull`, allow for registry authentication, pulling an image by URI, extracting and storing details by inspecting. In the following PRs these details will be used for populating the local cache and building the EIF with linuxkit.


enclave_build/src/image.rs
```diff
@@ -69,3 +69,183 @@ pub fn deserialize_from_reader<R: Read, T: DeserializeOwned>(reader: R) -> Resul
 
     Ok(deserialized_obj)
 }
+
+#[cfg(test)]
+mod tests {
```
lauralt, line 74: Unit tests should be in the same commit as the code they are testing, so that the user looking over a commit (this one for example) has full context on what's happening there.

lauralt (commented):
Thanks, @raulmldv! Code looks much cleaner now.

THIRD_PARTY_LICENSES_RUST_CRATES.html
```diff
@@ -44,10 +44,11 @@ <h1>Third Party Licenses</h1>
...
         <h2>Overview of licenses:</h2>
         <ul class="licenses-overview">
-            <li><a href="#MIT">MIT License</a> (29)</li>
-            <li><a href="#Apache-2.0">Apache License 2.0</a> (27)</li>
+            <li><a href="#Apache-2.0">Apache License 2.0</a> (36)</li>
+            <li><a href="#MIT">MIT License</a> (31)</li>
+            <li><a href="#BSD-3-Clause">BSD 3-Clause &quot;New&quot; or &quot;Revised&quot; License</a> (4)</li>
             <li><a href="#ISC">ISC License</a> (3)</li>
-            <li><a href="#BSD-3-Clause">BSD 3-Clause &quot;New&quot; or &quot;Revised&quot; License</a> (2)</li>
+            <li><a href="#Unicode-DFS-2016">Unicode License Agreement - Data Files and Software (2016)</a> (1)</li>
```
shtaked, line 51: Need to be double-checked form legal perspective

enclave_build/src/image.rs
```diff
@@ -0,0 +1,251 @@
...
+
+        Ok(Self {
+            uri: image_ref.whole(),
+            hash: image_hash,
+            config: deserialize_from_reader(image_data.config.data.as_slice())?,
+        })
+    }
+}
+
+/// For example, "hello-world" image has reference "docker.io/library/hello-world:latest".
```
shtaked, line 55: But it won't try to guess "hello-world" to "docker.io/library/hello-world:latest" according to source code from `oci_distribution` (https://github.com/krustlet/oci-distribution/blob/main/src/reference.rs#L181), will it?

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,171 @@
...
+pub const ACCEPTED_MEDIA_TYPES_COUNT: usize = 3;
+/// Accepted image layer media types.
+pub const ACCEPTED_MEDIA_TYPES: [&str; ACCEPTED_MEDIA_TYPES_COUNT] = [
+    oci_distribution::manifest::WASM_LAYER_MEDIA_TYPE,
+    oci_distribution::manifest::IMAGE_DOCKER_LAYER_GZIP_MEDIA_TYPE,
+    oci_distribution::manifest::IMAGE_LAYER_GZIP_MEDIA_TYPE,
+];
+/// Sequence of environment variables and config file paths to try parsing: (env, path)
+pub const CONFIG_ENV_PATHS: [(&str, &str); 3] = [
+    ("DOCKER_CONFIG", "/"),
```
shtaked, line 28: Wouldn't that ending to `/` look for a directory rather than file itself?

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,171 @@
...
+    let config_json: serde_json::Value = deserialize_from_reader(&config_file).map_err(|err| {
+        EnclaveBuildError::CredentialsError(format!("JSON was not well-formatted: {:?}", err))
+    })?;
+
+    let auths = config_json.get("auths").ok_or_else(|| {
+        EnclaveBuildError::CredentialsError("Could not find auths key in config JSON".to_string())
+    })?;
+
+    if let serde_json::Value::Object(auths) = auths {
+        for registry_auths in auths.values() {
```
shtaked, line 94: Previously in `docker.rs` we checked the registry name against a registry in a specified reference.
How will it work now? Shouldn't it be the same?

THIRD_PARTY_LICENSES_RUST_CRATES.html
raulmldv: This was auto-generated, will look into it.

enclave_build/Cargo.toml
```diff
@@ -23,3 +23,5 @@ futures = "0.3.13"
 thiserror = "1.0"
```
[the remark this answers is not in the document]
raulmldv: I guess the intention was to separate crates related to the image. Not really needed.

enclave_build/src/image.rs
raulmldv: To construct the reference, the `split_domain` function is used which will actually set the default domain to `docker.io` for those URIs that are missing one. If that's the case, it will complete the reference with `library` as well.

enclave_build/src/image.rs
```diff
@@ -69,3 +69,183 @@ pub fn deserialize_from_reader<R: Read, T: DeserializeOwned>(reader: R) -> Resul
...
+            config: config_obj,
+            manifest: Some(manifest_obj),
+        };
+
+        image_data
+    }
+
+    #[test]
+    fn test_from_image_data() {
+        // Use the mock image data from the cache.rs module
```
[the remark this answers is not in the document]
raulmldv: Moved the referenced data here, missed this comment.

enclave_build/src/pull.rs
raulmldv: Yes, path string should be empty.

enclave_build/src/pull.rs
raulmldv: This assumes that the only credential present is for the supplied URI. I guess it should check like in `docker.rs`

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,171 @@
...
+        .map_err(|err| EnclaveBuildError::ImagePullError(err.to_string()))?;
+
+    let auth = registry_auth();
+
+    // Pull from remote an ImageData struct containing the layers, manifest and configuration files
+    client
+        .pull(&image_ref, &auth, ACCEPTED_MEDIA_TYPES.to_vec())
+        .await
+        .map_err(|err| EnclaveBuildError::ImagePullError(err.to_string()))
+}
```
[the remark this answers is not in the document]
raulmldv: This is tested after adding the cache. I can add a test here as well.

enclave_build/src/image.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug)]
+/// Struct representing the image metadata, like the image ID (hash) and config.
+pub struct ImageDetails {
+    /// The reference of an image, e.g. "docker.io/library/hello-world:latest".
+    // Use a String, since the oci_distribution::Reference struct does not implement
+    // Serialize/Deserialize
+    uri: String,
+    /// The image ID, calculated as the SHA256 digest hash of the image config.
+    #[serde(rename = "Id")]
```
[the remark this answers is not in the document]
lauralt: Is the serde rename needed anymore?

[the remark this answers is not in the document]
raulmldv: Yes, later in the workflow it will try to match the `PascalCase` of the result returned by docker inspect.

shtaked (approved):

enclave_build/src/image.rs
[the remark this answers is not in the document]
raulmldv: The casing for the returned structure is declared here:
https://github.com/softprops/shiplift/blob/master/src/image.rs#L798

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,225 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use sha2::Digest;
+
+    /// Test pulled configuration hash is as expected
+    #[tokio::test]
+    async fn test_pull() {
```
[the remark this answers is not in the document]
raulmldv: A bit hard to test authentication functions as they rely on external scripts on the testing environment. I can add tests for credential search fallback and maybe an invalid URIs.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,282 @@
...
+        let user = "AWS";
+
+        let uri = SAMPLE_IMAGE;
+
+        let credentials = parse_credentials(uri).unwrap();
+
+        if let RegistryAuth::Basic(username, _password) = credentials {
+            assert_eq!(username, user);
+        } else {
+            assert!(false);
```
lauralt, line 253: as an FYI: you can use `unreachable!` instead of `assert!(false)`.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// The assumed format of the file is:\
+/// {\
+///        "auths": {\
+///            "https://index.docker.io/v1/": {\
+///                    "auth": "<token_string>"\
+///            }\
+///        }\
+/// }
+pub fn parse_credentials(image_name: &str) -> Result<RegistryAuth> {
```
lauralt, line 84: i think this doesn't need to be public.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,282 @@
...
+    ("XDG_RUNTIME_DIR", "containers/auth.json"),
+    ("HOME", ".docker/config.json"),
+];
+
+/// Builds a client which uses the protocol given as parameter.
+///
+/// Client required for the https://github.com/krustlet/oci-distribution library API.
+///
+/// By default, the client pulls the image matching the current running architecture.
+pub fn build_client(protocol: ClientProtocol) -> Client {
```
lauralt, line 37: does this need to be public?

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    // If the auth token is missing, return error
+    Err(EnclaveBuildError::CredentialsError(
+        "Credentials not found.".to_string(),
+    ))
+}
+
+/// Determines the authentication for interacting with the remote registry.
+pub fn registry_auth(image_name: &str) -> RegistryAuth {
```
lauralt, line 166: does this need to be public?

enclave_build/src/image.rs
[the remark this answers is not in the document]
lauralt: Thanks for managing to remove the serde rename!

enclave_build/src/pull.rs
raulmldv: None of these need to be public.

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,388 @@
...
+mod tests {
+    use std::{
+        io::{Read, Write},
+        path::PathBuf,
+    };
+
+    use super::*;
+    use serde_json::{json, Value};
+    use sha2::Digest;
+    use tempfile::{tempdir, TempDir};
```
lauralt, line 211: I know this is used in other places in this project as well, but it's an external dependency and we should try to avoid them when that's now too difficult to do. Can you please use [tempfile](https://github.com/rust-vmm/vmm-sys-util/blob/main/src/tempfile.rs) from vmm-sys-util? It shouldn't be too hard to switch to that.

We should also open an issue, to remove the tempfile dependency from the whole project.

shtaked (approved):

enclave_build/src/pull.rs
```diff
@@ -0,0 +1,384 @@
...
+        if let RegistryAuth::Basic(username, password) = credentials {
+            assert_eq!(username, TEST_USER);
+            assert_eq!(password, TEST_PASS);
+        } else {
+            unreachable!();
+        }
+    }
+
+    /// Test different credential parsing errors
+    fn test_parsing_failure(content: &Value) {
```
lauralt, line 306: nit: would be nice to rename helper functions so that they don't start with `test_`. It is confusing as you would expect this is a proper test, not something called by the actual tests.

lauralt (approved):

PR #423, opened 2022-11-21
Fix clippy errors and update driver-bindings

petreeftime:


lauralt (dismissed):

laurstefan (dismissed):

driver-bindings/src/bindings.rs
```diff
@@ -1,4 +1,4 @@
-/* automatically generated by rust-bindgen 0.59.2 */
+/* automatically generated by rust-bindgen 0.62.0 */
```
lauralt, line 1: Is [this](https://github.com/aws/aws-nitro-enclaves-cli/blob/main/Cargo.toml#L35) used anywhere? If not, we should probably remove it.

lauralt (approved):

driver-bindings/src/bindings.rs
petreeftime: It doesn't seem to be used anywhere.

petreeftime: I'll check if it can be removed in a future PR.

lauralt: Ok, thanks!

PR #424, opened 2022-11-22
enclave_build: fix argument parsing and image generation

sstone:
*Description of changes:*

- fixed inconsistencies in argument names that make it impossible to generate a signed image
- fixed potential image corruption if the output image already exists by truncating it before writing to it
- bump version to 0.2.0


sstone:
> Checked the logs: `Makefile:221: recipe for target 'nitro-format' failed` Make sure to run `cargo-fmt --all` before commiting.

done in https://github.com/aws/aws-nitro-enclaves-cli/pull/424/commits/dc6467971e9a9315a0ec64986b67be723567f741

raulmldv:
Currently failing on license check `make nitro-about`. Opened a PR to update our license file. After we merge it, you should rebase.

Cargo.lock
```diff
@@ -303,7 +303,7 @@ checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
 
 [[package]]
 name = "enclave_build"
-version = "0.1.0"
+version = "0.2.0"
```
lauralt, line 306: Why do we bump the version here? Is the version 0.1.0 released on crates.io?

Cargo.lock
```diff
@@ -303,7 +303,7 @@ checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
```
lauralt, line 303: Would be nice to follow [the 50/72 rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule) for commit messages. Also please add a `Signed-off-by` line: `git commit -s --amend`.

Cargo.lock
sstone: Since it's a significant change I thought it would make sense to change the version here, but I don't know how this library's maintainers manage versioning.

Cargo.lock
sstone: I'll rebase on `main` and update the commit message.

Cargo.lock
lauralt: If it's not on crates.io, I don't see a point in updating the version here.

sstone: I have no idea, I think it's up to the maintainers to choose here.

petreeftime:
Can you describe a bit more the issue that this patch is fixing? It seems to make the commandline arguments inconsistent, since they follow `snake_case` rather than `kebab-case` in `enclave_build` crate. I think they would all need to be updated to look the same.

sstone:
> Can you describe a bit more the issue that this patch is fixing? It seems to make the commandline arguments inconsistent, since they follow `snake_case` rather than `kebab-case` in `enclave_build` crate. I think they would all need to be updated to look the same.

The issue here is that it is not possible to create a signed image with `enclave_build` because the private key parameter is not parsed at all (instead the code reuses `private_certificate` which is imho a bug). I used `private-key` and not `private_key` because that's how the command line option is defined as the beginning of `main.rs`.

PR #425, opened 2022-11-28
ci: add workflows build, clippy and format workflows

petreeftime:
Currently all the workflows are being run in a private environment and the results from this workflow are not being published, which makes it difficult for contributors to understand and fix the issues they might encounter. This adds a separate CI for build, format and clippy, which should add at least some observability. Some of the tests require a specific environment to run, so currently they cannot be enabled to run in the github workflow, I will filter out these tests separately and either fix them, where possible, or filter them out from the list of tests to run in the workflow in a future PR.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,3 +39,13 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - run: rustup component add rustfmt
     - run: cargo fmt --all -- --check
+
+  audit:
+    name: Check dependencies for security issues
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


shtaked (dismissed):

petreeftime:
The failure is expected, requires #427 to pass, but it's not a blocking failure and we can make all the other checks mandatory, except this one. This would still give a signal that an update is required, but it only really needs to be updated on release.

raulmldv (approved):

shtaked (approved):

PR #426, opened 2022-11-28
ci: only run notifications for aws repository

petreeftime:

*Issue #, if available:*

*Description of changes:*


shtaked (approved):

.github/workflows/event.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 jobs:
   build:
+    if: ${{ github.repository == "aws/aws-nitro-enclaves-cli" }}
```
lauralt, line 12: Why do we need this if we have only one repository here? Can you explain in the commit message the purpose of this update?

shtaked: I guess it's to avoid running the workflow in forks, but yes, would be great to have it in the commit message.

mariusknaust (approved):

belveryin (approved):

PR #427, opened 2022-11-29
Update THIRD_PARTY_LICENSES_RUST_CRATES.html

raulmldv:
Fixed failing `nitro-about` target on recent PRs.

Signed-off-by: Raul-Ovidiu Moldovan <raulmldv@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/THIRD_PARTY_LICENSES_RUST_CRATES.html
+++ b/THIRD_PARTY_LICENSES_RUST_CRATES.html
@@ -44,8 +44,8 @@
     
         <h2>Overview of licenses:</h2>
         <ul class="licenses-overview">
+            <li><a href="#Apache-2.0">Apache License 2.0</a> (31)</li>
             <li><a href="#MIT">MIT License</a> (29)</li>
-            <li><a href="#Apache-2.0">Apache License 2.0</a> (27)</li>
             <li><a href="#ISC">ISC License</a> (3)</li>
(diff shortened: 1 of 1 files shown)


shtaked (approved):

petreeftime:
I think we should simply remove nitro-about from the default test set, or keep just a check for valid licenses. We only need to include this file when we produce a binary distribution.

raulmldv:
> I think we should simply remove nitro-about from the default test set, or keep just a check for valid licenses. We only need to include this file when we produce a binary distribution.

Couldn't find a quick check using `cargo-about` or `cargo-license`. I will remove `nitro-about` for now and look more into it. I think I can leave the licenses updated so I'll push here.

Makefile
```diff
@@ -241,17 +241,6 @@ nitro-audit: build-setup build-container
 			'source /root/.cargo/env && \
 			cargo audit -f /nitro_src/Cargo.lock --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071'
 
-nitro-about: build-setup build-container
```
petreeftime, line 244: Why are you removing this?

raulmldv: Right, I can just remove it from the script.

petreeftime: Let's merge #425 first.

PR #429, opened 2022-12-05
OCI storage

raulmldv:
OCI storage

Implemented an OCI compliant storage for pull and inspect operations. This storage can be directly used by `linuxkit` as its own cache, preventing it to pull the images again. The structure of the storage directory is as follows:

```
storage_dir/
|________index.json        -> points to the manifest blob for each image
|________oci-layout
|            blobs/sha256/
|            |_______manifest_hash  -> points to config and layers
|            |_______config_hash
|            |_______layer_hash
```
The storage module supports store and fetching operations. The first fetch validates the files for the requested image and loads the config in the memory (the following fetches will use this config). Only one config is stored as we do not need to keep track of all the stored images when building an EIF.

When later adding the OCI archive feature, extracted image contents can be directly added to the storage and used by `nitro-cli` and `linuxkit` as well.

### DIFF
--- a/enclave_build/docs/storage.md
+++ b/enclave_build/docs/storage.md
@@ -1,0 +1,171 @@
+# Image Local Storage
+
+When building an enclave image using the Docker daemon, the image is pulled and stored in the
+Docker storage (place in the filesystem where images pulled from remote registries are saved locally).
+We want to implement the same functionality when choosing to build an OCI image.
+The resulted storage follows the [OCI specification](https://github.com/opencontainers/image-spec),
+allowing tools like `linuxkit` to use it as its own cache, and also adding images from OCI archive
+with the same format.
+
+To ease the interactions with the image details, the storage structure holds the config of the
+image used to build the EIF. This prevents unnecessary file reads.
+
+## Structure
+
+The root folder of the storage is either `${XDG_DATA_HOME}/.aws-nitro-enclaves-cli/container_storage`,
+or, if `XDG_DATA_HOME` is not set, `${HOME}/.local/share/.aws-nitro-enclaves-cli/container_storage`
+is used.
+
+The storage holds common components used by all images (index file, layout file, blobs folder) and
+individual items specific to the image named blobs (manifest, config file, layers). A possible
+storage structure can look like this:
+
+```
+storage_dir/
+|________index.json   -> points to the manifest blob file for each image (maps image reference to manifest digest)
+|________oci-layout
+|            blobs/sha256/
+|            |_______98806831de537d56f19b4b3cbf6ed80187374a1521a18d2d0c5f8a0e3962b2ae -> manifest_hash points to config and layers
+|            |_______feb5d9fea6a5e9606aa995e879d862b825965ba48de054caab5ef356dc6b3412 -> config_hash
+|            |_______2db29710123e3e53a794f2694094b9b4338aa9ee5c40b930cb8063a1be392c54 -> layer_hash
+```
+
+### Examples
+
+To better understand the storage structure and how the index and manifest point to other files, we
+can follow a `hello-world` image example.
+
+#### Index
+
+index.json
+```
+{
+  "schemaVersion": 2,
+  "manifests": [
+    {
+      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+      "size": 425,
+      "digest": "sha256:98806831de537d56f19b4b3cbf6ed80187374a1521a18d2d0c5f8a0e3962b2ae",
+      "annotations": {
+        "org.opencontainers.image.ref.name": "docker.io/test-oci:hello"
+      }
+    }
+  ]
+}
+```
+
+As we can see, the index holds one manifest entry, for one image stored. The number of manifests
+increases with each newly stored image. The manifest hash points us to the manifest blob.
+
+#### Layout file
+
+oci-layout:
+```
+{"imageLayoutVersion":"1.0.0"}
+```
+
+This file simply defines the OCI image layout version we are using.
+
+#### Manifest
+
+98806831de537d56f19b4b3cbf6ed80187374a1521a18d2d0c5f8a0e3962b2ae
+```
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+  "config": {
+    "mediaType": "application/vnd.docker.container.image.v1+json",
+    "digest": "sha256:feb5d9fea6a5e9606aa995e879d862b825965ba48de054caab5ef356dc6b3412",
+    "size": 1469
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+      "digest": "sha256:2db29710123e3e53a794f2694094b9b4338aa9ee5c40b930cb8063a1be392c54",
+      "size": 2479
+    }
+  ]
+}
+```
+
+The manifest is our guide for a particular image. It tells us which components among the blobs
+belong to the image it defines. It tells us the digest of the content for the config file and each
+layer. With this information we can search the blobs.
+
+#### Config
+
+The configuration file defines the behavior of the image inside the container, platform
+dependencies and interaction with the filesystem composed from the layers. Nitro CLI's usage of
+the storage should not be concerned with the contents of this file (excepting UTF-8 and digest
+validation).
+
+#### Layers
+
+Layers are serialized filesystem changes that go on top of each other, starting from the base
+filesystem we create the image with. The file has binary content that, from Nitro CLI's point of
+view, just has to obey the content digest in the manifest.
+
+## Operations
+
+The operations offered by the `storage` module are as follows:
+
+### Create
+
+When building our first image, the storage does not exist. This operation creates, prior to
+storing an image, an empty storage. It contains the storage root folder and a blobs folder. Also
+at this point, the config of the current image is not yet loaded in the memory.
+
+### Store
+
+This is the operation which adds an image to the storage. It is called on an empty storage or after
+a fetch fails when searching for the image. After pulling the image and extracting its details, its
+manifest entry is added to the index and the blobs are written, each to their file.
+
+### Fetch details
+
+This operation returns the details of an image (image name, config digest and the config itself).
+These details are saved in the memory as the config of the image we are currently using to build
+the EIF. If a fetch was performed before, the config is already loaded and we can just return it.
+No other file read is needed.
+
+When we fetch an image without the config already loaded, we want to also check that it is stored
+correctly (or stored at all). The validation starts from the index file. It should contain our
+image's reference and the digest of the manifest. If fields are missing or the JSON structure does
+not comply, an error is returned.
+
+The validation continues with the manifest. As we did with the index, we validate the structure and
+get the digests for the config and blobs to later check them as well. We also check that the
+content of the manifest matches its digest from the file name.
+
+The config validation simply checks that a file with its content digest exists and the content is
+UTF-8 formatted. The file name is also validated against the content hash.
+
+Having a list of the layers' digest from the manifest, we go through each of them to get access to
+the blob files. By hashing each of the files and comparing the result with the initial digest, we
+validate the layers as well.
+
+
+### Get root folder
+
+So we can get access to the storage contents outside the `storage` module (in our case for
+`linuxkit`), we want to be able to return the location of the storage files. The module returns
+the root folder of the storage, as decided by the values of the environment variables
+`XDG_DATA_HOME` or `HOME`.
+
+## Workflow
+
+To see how these operations fit with each other and the structure as well, let's explore the common
+workflow:
+
+1. Given an image name we want to build an EIF
+2. Perform the first fetch:
+  - Check if the config of the current image is loaded. If it is, just return the image details
+  built from this config.
+  - If there is no config, read the files and perform the validation while searching for the
+  image files.
+  - If the validation hasn't failed, return the config.
+3. If the fetch failed (storage empty, missing our image, or malformed), pull the image from the
+remote registry, store its contents and load the config.
+4. A second fetch (for the inspect) returns the image details using the config from the memory
+without additional steps.
+
--- a/enclave_build/src/lib.rs
+++ b/enclave_build/src/lib.rs
@@ -66,6 +66,8 @@ pub enum EnclaveBuildError {
     OciStorageStore(std::io::Error),
     #[error("Storage entry not found or has wrong: `{0:?}`")]
     OciStorageNotFound(String),
+    #[error("Storage has the wrong structure: `{0:?}`")]
+    OciStorageMalformed(String),
     #[error("Manifest missing or wrong format")]
     ConfigError,
     #[error("Manifest missing or wrong format")]
--- a/enclave_build/src/lib.rs
+++ b/enclave_build/src/lib.rs
@@ -9,6 +9,7 @@ use std::process::Command;
 mod docker;
 mod image;
 mod pull;
+mod storage;
 mod yaml_generator;
 
 use aws_nitro_enclaves_image_format::defs::{EifBuildInfo, EifIdentityInfo, EIF_HDR_ARCH_ARM64};
@@ -59,12 +60,12 @@ pub enum EnclaveBuildError {
     SerdeError(serde_json::Error),
     #[error("Error while parsing credentials: `{0}`")]
     CredentialsError(String),
-    #[error("Image cache initialization error: `{0:?}`")]
-    CacheInitError(std::io::Error),
-    #[error("Cache store operation failed: `{0:?}`")]
-    CacheStoreError(std::io::Error),
-    #[error("Cache entry not found or has wrong: `{0:?}`")]
-    CacheMissError(String),
+    #[error("Image storage initialization error: `{0:?}`")]
+    OciStorageInit(std::io::Error),
+    #[error("Image store operation failed: `{0:?}`")]
+    OciStorageStore(std::io::Error),
+    #[error("Storage entry not found or has wrong: `{0:?}`")]
+    OciStorageNotFound(String),
     #[error("Manifest missing or wrong format")]
     ConfigError,
     #[error("Manifest missing or wrong format")]
--- a/enclave_build/src/storage.rs
+++ b/enclave_build/src/storage.rs
@@ -428,6 +428,40 @@ impl OciStorage {
             image_config.clone(),
         ))
     }
+
+    /// Returns the default root folder path of the local storage.
+    ///
+    /// The default storage path is {XDG_DATA_HOME}/.aws-nitro-enclaves-cli/container_storage, and if the env
+    /// variable is not set, {HOME}/.local/share/.aws-nitro-enclaves-cli/container_storage is used.
+    pub fn get_default_root_path() -> Result<PathBuf> {
+        // Try to use XDG_DATA_HOME as default root
+        let root = match std::env::var_os(STORAGE_ROOT_FOLDER) {
+            Some(val) => val
+                .into_string()
+                .map_err(|_| EnclaveBuildError::PathError("storage root folder".to_string()))?,
+            // If XDG_DATA_HOME is not set, use {HOME}/.local/share as specified in
+            // https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+            None => {
+                let home_folder = std::env::var_os("HOME").ok_or_else(|| {
+                    EnclaveBuildError::PathError(
+                        "HOME environment variable is not set.".to_string(),
+                    )
+                })?;
+                format!(
+                    "{}/.local/share/",
+                    home_folder
+                        .into_string()
+                        .map_err(|err| EnclaveBuildError::PathError(format!("{:?}", err)))?
+                )
+            }
+        };
+
+        let mut path = PathBuf::from(root);
+        // Add the additional path to the root
+        path.push(".aws-nitro-enclaves-cli/container_storage");
+
+        Ok(path)
+    }
 }
 
 #[cfg(test)]
@@ -629,5 +663,25 @@ pub mod tests {
         assert!(storage
             .fetch_image_details(image::tests::TEST_IMAGE_NAME)
             .is_err());
+    }
+
+    #[test]
+    fn test_get_default_root_path() {
+        let default_path = match std::env::var_os("XDG_DATA_HOME") {
+            Some(val) => PathBuf::from(val.to_str().unwrap().to_string())
+                .join(".aws-nitro-enclaves-cli/container_storage"),
+            None => {
+                let home = std::env::var_os("HOME").expect("HOME env not set");
+                let append = format!(
+                    "{}/.local/share/.aws-nitro-enclaves-cli/container_storage",
+                    home.to_str().unwrap().to_string()
+                );
+                PathBuf::from(home).join(append)
+            }
+        };
+        let calc_path =
+            OciStorage::get_default_root_path().expect("failed to determine default storage path");
+
+        assert_eq!(default_path, calc_path);
     }
 }
--- a/enclave_build/src/storage.rs
+++ b/enclave_build/src/storage.rs
@@ -40,6 +40,8 @@ const OCI_LAYOUT: (&str, &str) = ("imageLayoutVersion", "1.0.0");
 /// Structure which provides operations with the local storage.
 ///
 /// The index file is located in the storage root folder and keeps track of stored images.
+/// The storage will be consumed during image operations (pull and inspect) to prevent requests to the
+/// image registry. Also `linuxkit` will use this storage as its own to prevent subsequent requests
 ///
 /// The storage structure is:
 ///
@@ -75,7 +77,7 @@ impl OciStorage {
     }
 
     /// Stores the image data provided as argument in the storage at the folder pointed
-    /// by the 'root_path' field.
+    /// by the 'root_path' field: adds image to the index content, stores manifest, config, layers
     pub fn store_image_data(&mut self, image_name: &str, image_data: &ImageData) -> Result<()> {
         // Create the folder where the image data will be stored. Each image blob will be stored in
         // a file named by the SHA256 digest of the content.
@@ -110,6 +112,9 @@ impl OciStorage {
         }
     }
 
+    // Each layer file will be named after the layer's digest hash. Only blobs found in the
+    // image manifest `layer` section:
+    // https://github.com/opencontainers/image-spec/blob/main/image-layout.md#example-blobs
     fn store_layers(blobs_path: &Path, image_data: &ImageData) -> Result<()> {
         for layer in &image_data.layers {
             Self::write_blob(blobs_path, &layer.data)?;
@@ -118,6 +123,8 @@ impl OciStorage {
         Ok(())
     }
 
+    // Store the manifest as a blob with the following content:
+    // https://github.com/opencontainers/image-spec/blob/main/manifest.md#example-image-manifest
     fn store_manifest<'a>(
         blobs_path: &Path,
         image_data: &'a ImageData,
@@ -133,6 +140,9 @@ impl OciStorage {
         Ok(manifest)
     }
 
+    // Store the config and validate UTF8 bytes. Saved as a blob as well
+    // with the hash saved in the manifest:
+    // https://github.com/opencontainers/image-spec/blob/main/config.md#example
     fn store_config(blobs_path: &Path, image_data: &ImageData) -> Result<String> {
         let config_content = String::from_utf8(image_data.config.data.clone()).map_err(|_| {
             EnclaveBuildError::OciStorageStore(Error::new(
@@ -177,7 +187,8 @@ impl OciStorage {
             .open(self.root_path.join(STORAGE_INDEX_FILE_NAME))
             .map_err(EnclaveBuildError::OciStorageStore)?;
 
-        // Write index file content
+        // Write index file content. It must be present and defines the OCI storage structure:
+        // https://github.com/opencontainers/image-spec/blob/main/image-layout.md#oci-layout-file
         serde_json::to_writer(index_file, &index_content).map_err(EnclaveBuildError::SerdeError)?;
 
         Ok(())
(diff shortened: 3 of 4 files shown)


enclave_build/src/cache.rs
```diff
@@ -0,0 +1,220 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// S&PDX-License-Identifier: Apache-2.0
```
petreeftime, line 2: Typo: Should be SPDX

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// If an index file exists at the path, it loads the file's contents into the 'cached_images'
+    /// field. If not, a new empty index file is created at that path.
+    pub fn new(root_path: &Path) -> Result<Self> {
+        // Create all missing folders, if not already created
+        fs::create_dir_all(&root_path).map_err(EnclaveBuildError::CacheInitError)?;
+
+        // Try to open the index file and read the contents.
+        // If the file is missing, create it.
+        OpenOptions::new()
+            .write(true)
```
petreeftime, line 112: Why is read required here?

enclave_build/src/cache.rs
```diff
@@ -242,6 +242,145 @@ impl CacheManager {
...
+            // The 'cached_layers' hashmap will contain all layer files found in the cache for the current image
+            .for_each(|(name, file)| {
+                if let Ok(file_name) = name.into_string() {
+                    cached_blobs.insert(file_name, file);
+                }
+            });
+
+        // Iterate through each layer found in the image manifest and validate that it is stored in
+        // the cache by checking the digest
+        for layer_obj in layers_vec {
```
petreeftime, line 330: This should be split into a separate function to keep this function simple and then you can also use a `map` or `for_each`.

enclave_build/src/cache.rs
```diff
@@ -242,6 +242,145 @@ impl CacheManager {
...
+                })?
+                .to_string();
+
+            // Get the cached layer file matching the digest
+            // If not present, then a layer file is missing, so return Error
+            let mut layer_file = cached_blobs.get(&layer_digest).ok_or_else(|| {
+                EnclaveBuildError::CacheMissError("Layer missing from cache.".to_string())
+            })?;
+            let mut layer_bytes = Vec::new();
+            layer_file.read_to_end(&mut layer_bytes).map_err(|_| {
```
petreeftime, line 355: Would `read_to_string` make more sense here?

enclave_build/src/cache.rs
```diff
@@ -467,6 +470,39 @@ impl CacheManager {
...
+    ///
+    /// If the data is not correctly cached or a file is missing, it returns an error.
+    ///
+    /// If the image is not cached, it does not attempt to pull the image from remote.
+    pub fn fetch_image_details(&self, image_name: &str) -> Result<ImageDetails> {
+        let config_hash = self
+            .get_image_hash_from_name(&Self::normalize_reference(image_name))
+            .ok_or_else(|| EnclaveBuildError::CacheMissError("Image hash missing".to_string()))?;
+
+        // // Add algorithm prefix to the hash
```
petreeftime, line 484: nit: Double `//`.

petreeftime (commented):
In general the code looks correct, however, I found it difficult to read the code. A lot of the comments say what the code is doing, but I would prefer to also have some comments about why it does this (ie. either include documentation about how this image cache looks like inline, or provide links to the specific documentation that is relevant).

enclave_build/src/cache.rs
raulmldv: Missed it during clean up.

shtaked (commented):
Generally looks good, the only thing, as I mentioned I would leverage `oci-distribution` structures for reading/storing the content of the cache.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,219 @@
...
+pub const CACHE_OCI_LAYOUT_FILE: &str = "oci-layout";
+
+/// Constants used for complying with the OCI cache structure
+pub const DEFAULT_MEDIA_TYPE: &str = "application/vnd.oci.image.manifest.v1+json";
+pub const REF_ANNOTATION: &str = "org.opencontainers.image.ref.name";
+pub const OCI_LAYOUT: (&str, &str) = ("imageLayoutVersion", "1.0.0");
+
+/// Manifest entry in index file
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ManifestEntry {
```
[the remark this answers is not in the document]
raulmldv: I tried to match the `linuxkit` format of the index file, but I see that `oci-distirbution` has a more general structure which might work. This could also loosen the coupling with `linuxkit`.

enclave_build/src/cache.rs
```diff
@@ -201,13 +300,110 @@ impl CacheManager {
...
     }
+
+    /// Add `docker.io` to references that are missing this registry so `linuxkit` can validate cache presence
+    fn normalize_reference(reference: &str) -> String {
+        let docker_prefix = "docker.io/";
+        if reference.starts_with(docker_prefix) {
+            return reference.to_string();
+        }
+
+        docker_prefix.to_owned() + reference
```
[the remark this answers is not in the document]
raulmldv: So `linuxkit` will take the image reference and automatically append `docker.io/` to search for it in the cache and will not recognize `public.ecr.aws/` as an already existing registry. If I don't save the image like this, `linuxkit` will search an image reference that starts with `docker.io/public.ecr.aws/` and will fail to find it in the cache and later to pull it.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,219 @@
...
+impl CacheManager {
+    /// Creates a new CacheManager instance and returns it. As argument, a path to the root folder
+    /// of the cache should be provided.
+    ///
+    /// Apart from that, the function also creates (if not already created) all folders from the path
+    /// specified as argument.
+    ///
+    /// If an index file exists at the path, it loads the file's contents into the 'cached_images'
+    /// field. If not, a new empty index file is created at that path.
+    pub fn new(root_path: &Path) -> Result<Self> {
```
lauralt, line 111: Does it help to use `Path` instead of `PathBuf`? We would avoid all the `to_path_buf`s.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,219 @@
...
+        Ok(Self {
+            root_path: root_path.to_path_buf(),
+            cached_images,
+        })
+    }
+
+    fn fetch_index(root_path: &Path) -> Result<CacheIndex> {
+        match File::options()
+            .read(true)
+            .open(root_path.join(CACHE_INDEX_FILE_NAME))
```
lauralt, line 410: Can't we use `File::open` directly?

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,219 @@
...
+                .filter_map(|entry| {
+                    entry
+                        .annotations
+                        .get(REF_ANNOTATION)
+                        .map(|img_ref| (img_ref.to_string(), entry.digest.to_string()))
+                })
+                // Read manifest files and map digest to JSON content
+                .filter_map(|(img_ref, manifest_hash)| {
+                    Self::fetch_manifest(root_path, &manifest_hash)
+                        .ok()
```
lauralt, line 139: So if the `fetch_*` functions return an error, we discard it or what is happening here?

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,219 @@
...
+            .as_str()
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "Failed to get config digest from image manifest.".to_string(),
+                )
+            })?
+            .strip_prefix("sha256:")
+            .ok_or_else(|| {
+                EnclaveBuildError::CacheMissError(
+                    "Failed to get config digest from image manifest.".to_string(),
```
lauralt, line 477: Can we give a different error message here than the one above?

enclave_build/src/cache.rs
```diff
@@ -143,6 +146,102 @@ impl CacheManager {
...
+        })?;
+        let config_digest = format!("{:x}", sha2::Sha256::digest(config_json.as_bytes()));
+
+        File::create(&blobs_path.join(&config_digest))
+            .map_err(EnclaveBuildError::CacheStoreError)?
+            .write_all(config_json.as_bytes())
+            .map_err(EnclaveBuildError::CacheStoreError)?;
+
+        // If index file present, read and append new image entry
+        let index_json: CacheIndex = match File::options().read(true).open(CACHE_INDEX_FILE_NAME) {
```
lauralt, line 208: Again I think you can use File::open, also I think you can make the `match` a bit more Rusty, by using `map_or_else` or something else.

enclave_build/src/cache.rs
```diff
@@ -143,6 +146,102 @@ impl CacheManager {
...
+            Ok(file) => serde_json::from_reader(file).map_err(EnclaveBuildError::SerdeError)?,
+            Err(_) => CacheIndex::default(),
+        };
+        let mut index_content = index_json;
```
lauralt, line 212: Do we really need a new variable here?

enclave_build/src/cache.rs
```diff
@@ -143,6 +146,102 @@ impl CacheManager {
         })
     }
 
+    /// Stores the image data provided as argument in the cache at the folder pointed
+    /// by the 'root_path' field.
+    pub fn store_image_data(&mut self, image_name: &str, image_data: &ImageData) -> Result<()> {
```
lauralt, line 160: This method is way too big, and complicated I would say? Can you think of some ways to simplify it, or maybe split it in multiple functions etc.?

lauralt (commented):
First part of review.

enclave_build/src/cache.rs
raulmldv: No, missed it.

enclave_build/src/cache.rs
```diff
@@ -0,0 +1,764 @@
...
+/// {CACHE_ROOT_PATH}/blobs/sha256/hash_of_blob_2\
+/// etc.
+///
+/// A blob can be:
+///
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/manifest_hash - the image manifest, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/config_hash - the image configuration, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/layer_hash - one of the image layers, each in a separate gzip compressed tar file.
+#[derive(Clone)]
+pub struct CacheManager {
```
lauralt, line 93: I feel we have a design problem here. For example we are doing the same operations multiple times, e.g. instead of fetching the cache index file multiple times, we should do that just once, and keep it in the cache manager (as a separate field).
I'll try to think of other ways of how we can improve the cache manager abstraction.

shtaked (approved):

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
+# Image Cache
+
+When building an enclave image using the Docker daemon, the image will be pulled and stored in the Docker cache. We want to implement the same functionality when choosing to build an OCI image. The resulted cache will follow the [OCI specification](https://github.com/opencontainers/image-spec), allowing tools like `linuxkit` to use it as its own cache, and also adding images from OCI archive with the same format.
```
lauralt, line 3: Please limit the lines in md files to 80 chars; this is the general recommendation, otherwise it is difficult to read such files in an IDE for example.

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+      "digest": "sha256:98806831de537d56f19b4b3cbf6ed80187374a1521a18d2d0c5f8a0e3962b2ae",
+      "annotations": {
+        "org.opencontainers.image.ref.name": "docker.io/test-oci:hello"
+      }
+    }
+  ]
+}
+```
+
+As we can see, the index holds one manifest entry, for one image stored. The number of manifests increases with each newly stored image. The manifest hash will point us to the manifest blob.
```
lauralt, line 45: As already mentioned, please avoid the use of `will`. Makes the doc hard to read.

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+### Get root folder
+
+So we can get access to the cache contents outside the `CacheManager` (in our case for `linuxkit`), we want to be able to return the location of the cache files. It will return the root folder of the cache, as decided by the values of the environment variables `XDG_DATA_HOME` or `HOME`.
+
+## Workflow
+
+To see how these operation fit with each other and the structure as well, let's explore the common workflow:
+
+1. Given an image name we want to build an EIF
+2. Check if the cache exists. If not, pull the image and store it to the new cache.
```
lauralt, line 123: Which new cache? Please be more explicit.

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+
+This operation will return the details of an image (image name, config digest and the config itself). These details are saved in the memory of the `CacheManager` and will hold the information needed to build the enclave image. This operation is called after validating the cache.
+
+### Get root folder
+
+So we can get access to the cache contents outside the `CacheManager` (in our case for `linuxkit`), we want to be able to return the location of the cache files. It will return the root folder of the cache, as decided by the values of the environment variables `XDG_DATA_HOME` or `HOME`.
+
+## Workflow
+
+To see how these operation fit with each other and the structure as well, let's explore the common workflow:
```
lauralt, line 120: operations*

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+
+When building our first image, the cache will probably not exist. This operation will create, prior to storing an image, an empty cache. It will contain the cache root folder, an index file, a blobs folder and the oci-layout.
+
+### Store
+
+This is the operation which adds an image to the cache. It will be called on an empty cache or after a cache miss when searching for the image. After pulling the image and extracting its details, its manifest entry will be added to the index and the blobs will be written, each to their file.
+
+### Validate
+
+After we find an image in the index or the in-memory cache, we want to validate its structure before using the data. The validation starts from the index file. It should contain our image's reference and the digest of the manifest. If fields are missing or the JSON structure does not comply, an error will be returned.
```
lauralt, line 102: Is the in-memory cache the same you are talking about until this point or are we discussing about two different caches?

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+So we can get access to the cache contents outside the `CacheManager` (in our case for `linuxkit`), we want to be able to return the location of the cache files. It will return the root folder of the cache, as decided by the values of the environment variables `XDG_DATA_HOME` or `HOME`.
+
+## Workflow
+
+To see how these operation fit with each other and the structure as well, let's explore the common workflow:
+
+1. Given an image name we want to build an EIF
+2. Check if the cache exists. If not, pull the image and store it to the new cache.
+3. Check if the image is in the cache (in-memory). If not, pull the image store add it to the cache.
+4. Validate the image's files in the cache. If it fails, pull the image and overwrite cache.
```
lauralt, line 125:
```suggestion
4. Validate the image's files in the cache. If the validation fails, pull the image and overwrite cache.
```

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+
+So we can get access to the cache contents outside the `CacheManager` (in our case for `linuxkit`), we want to be able to return the location of the cache files. It will return the root folder of the cache, as decided by the values of the environment variables `XDG_DATA_HOME` or `HOME`.
+
+## Workflow
+
+To see how these operation fit with each other and the structure as well, let's explore the common workflow:
+
+1. Given an image name we want to build an EIF
+2. Check if the cache exists. If not, pull the image and store it to the new cache.
+3. Check if the image is in the cache (in-memory). If not, pull the image store add it to the cache.
```
lauralt, line 124: The second sentence might be missing some words.

enclave_build/docs/cache.md
lauralt: Overwrite cache with what? Do we remove the image or what are we doing?

enclave_build/docs/cache.md
```diff
@@ -0,0 +1,127 @@
...
+
+Layers are serialized filesystem changes that go on top of each other, starting from the base filesystem we create the image with. The file will be binary content that, from Nitro CLI's point of view, just has to obey the content digest in the manifest.
+
+## Operations
+
+The operations offered by the `cache` module are as follows:
+
+### Create
+
+When building our first image, the cache will probably not exist. This operation will create, prior to storing an image, an empty cache. It will contain the cache root folder, an index file, a blobs folder and the oci-layout.
```
lauralt, line 94: The files in the empty cache will have a default, initial state, or are they going to be empty files? Please add details about this.

lauralt (commented):
Thanks for adding documentation. Left some feedback for it.

enclave_build/docs/cache.md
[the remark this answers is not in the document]
raulmldv: As it is a simplified version of the specification (does not contain all the possible fields), I thought an example would make it easier to understand.

enclave_build/docs/cache.md
raulmldv: The cache itself is made up of the file structure presented. The in-memory cache is what the `CacheManager` loads (the map: image reference -> config digest). I will clarify.

enclave_build/docs/cache.md
raulmldv: Write the new content of the invalid files and truncate the file.

enclave_build/docs/cache.md
petreeftime: I think rust-fmt is 100 or 120 chars/line and I think it's fine if we keep a limit that works on a modern PC. But the line needs to be split for sure.

lauralt: Yes, I don't have a strong preference for 80, mostly just for splitting the lines.

raulmldv:
Refactored the implementation from a cache to a storage. Following our discussions, I came to the conclusion that we do not need to store config information in memory for all the images present in the index. We just need the one image used to build the EIF when running `enclave-build`. This behaviour does not resemble a cache, thus the renaming and documentation changes.

To avoid file reads for any storage access and loading information for images we do not use, a singe config is kept in memory and is loaded on the first fetch operation. The first fetch performs the files reads and validations and the next ones will just use the config to build the image details. Storing an image (after pulling from a remote registry) automatically loads the config without validations.

enclave_build/src/storage.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/manifest_hash - the image manifest, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/config_hash - the image configuration, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/layer_hash - one of the image layers, each in a separate gzip compressed tar file.
+#[derive(Clone)]
+pub struct OciStorage {
+    /// The root folder of the storage
+    root_path: PathBuf,
+
+    /// The config of the OCI image used to build the EIF
```
lauralt, line 64: You should explain why this is optional.

enclave_build/src/storage.rs
```diff
@@ -0,0 +1,62 @@
...
+    fs::{self},
+    path::{Path, PathBuf},
+};
+
+use oci_spec::image::ImageConfiguration;
+
+use crate::{EnclaveBuildError, Result};
+
+/// Root folder for the image storage.
+pub const STORAGE_ROOT_FOLDER: &str = "XDG_DATA_HOME";
```
lauralt, line 28: Do we need these as pub?

enclave_build/src/storage.rs
```diff
@@ -59,4 +71,312 @@ impl OciStorage {
...
+    pub fn store_image_data(&mut self, image_name: &str, image_data: &ImageData) -> Result<()> {
+        // Create the folder where the image data will be stored. Each image blob will be stored in
+        // a file named by the SHA256 digest of the content.
+        let blobs_path = self.root_path.join(STORAGE_BLOBS_FOLDER);
+        fs::create_dir_all(&blobs_path).map_err(EnclaveBuildError::OciStorageStoreError)?;
+
+        // Each layer file will be named after the layer's digest hash
+        Self::store_layers(&blobs_path, image_data)?;
+
+        // Store the manifest
```
lauralt, line 90: This comment is not useful.

enclave_build/src/storage.rs
```diff
@@ -59,4 +71,312 @@ impl OciStorage {
...
+
+        // Write oci_layout file from template constant
+        self.store_layout()?;
+
+        self.config = Some(deserialize_from_reader(config_content.as_bytes())?);
+
+        Ok(())
+    }
+
+    fn default_oci_index() -> OciImageIndex {
```
lauralt, line 107: Can't we implement Default instead?

enclave_build/src/storage.rs
```diff
@@ -0,0 +1,62 @@
...
+/// {STORAGE_ROOT_PATH}/blobs/sha256/hash_of_blob_1\
+/// {STORAGE_ROOT_PATH}/blobs/sha256/hash_of_blob_2\
+/// etc.
+///
+/// A blob can be:
+///
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/manifest_hash - the image manifest, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/config_hash - the image configuration, stored as a JSON String.\
+/// {IMAGE_FOLDER_PATH}/blobs/sha256/layer_hash - one of the image layers, each in a separate gzip compressed tar file.
+#[derive(Clone)]
```
lauralt, line 59: Is the Clone needed?

enclave_build/src/lib.rs
```diff
@@ -59,12 +60,12 @@ pub enum EnclaveBuildError {
...
     #[error("Error while parsing credentials: `{0}`")]
     CredentialsError(String),
-    #[error("Image cache initialization error: `{0:?}`")]
-    CacheInitError(std::io::Error),
-    #[error("Cache store operation failed: `{0:?}`")]
-    CacheStoreError(std::io::Error),
-    #[error("Cache entry not found or has wrong: `{0:?}`")]
-    CacheMissError(String),
+    #[error("Image storage initialization error: `{0:?}`")]
+    OciStorageInitError(std::io::Error),
```
lauralt, line 64: I've just noticed, you don't have to include `Error` in the variant name, it is clear that it's an error because it is part of an `Error` enum. Even if you only add other variants here, let's not propagate the issue, and name the new ones (or the changed ones) without the `Error` part.

enclave_build/src/storage.rs
```diff
@@ -59,4 +71,312 @@ impl OciStorage {
...
+    use sha2::Digest;
+    use std::env::temp_dir;
+    use vmm_sys_util::rand;
+
+    /// This function stores the test image in a temporary directory and returns that directory and
+    /// the storage manager initialized with it as root path.
+    fn setup_temp_storage() -> (PathBuf, OciStorage) {
+        // Use a temporary dir as the storage root path
+        // Create temporary random path so tests running in parallel won't overlap
+        let root_dir = temp_dir().join(rand::rand_alphanumerics(5));
```
lauralt, line 493: isn't it easier to use `TempDir` from vmm-sys-util?

enclave_build/src/lib.rs
raulmldv: I can remove `Error` for the ones that express a problem like `OciStorageNotFound` or `OciStorageMalformed`. The others feel weird without naming them failures or errors. This should be consistent with the naming throughout the repo.

enclave_build/src/storage.rs
raulmldv: Might be used in the future as public, will remove it for now.

enclave_build/src/storage.rs
raulmldv: I can remove it for now.

enclave_build/src/storage.rs
raulmldv: This is the default index we use in this storage, not the default of the structure itself.

enclave_build/src/lib.rs
lauralt: Hmm I don't necessarily find it weird, for example if you have `Error::OciStorageInit` you would read this as: we have an error in the oci storage initialization. This is one of the consequences of [namespacing](https://en.wikipedia.org/wiki/Namespace), i.e. the `Error` part is simply redundant. If the name sounds weird maybe it is missing some verb, but the Error part should not be the one that is missing.

raulmldv: Alright, makes sense. Changed the new ones.

enclave_build/src/storage.rs
```diff
@@ -398,10 +398,38 @@ impl OciStorage {
...
+        }
+        let image_config = self.config.as_ref().ok_or(EnclaveBuildError::ConfigError)?;
+
+        // Add algorithm prefix to the hash
+        let image_hash = format!(
+            "sha256:{:x}",
+            sha2::Sha256::digest(
+                &serde_json::to_vec(&self.config).map_err(EnclaveBuildError::SerdeError)?
+            )
+        );
```
lauralt, line 431: I'm not sure if I suggested at some point to not have this as a separate function, probably I did that, sorry, but looking again how many times this is used, I think it would be better though to define one, so that in case we change the algorithm we won't need to update it in multiple places.

enclave_build/src/storage.rs
```diff
@@ -398,10 +398,38 @@ impl OciStorage {
...
+    /// Fetches the image metadata from storage as an ImageDetails struct.
+    ///
+    /// If the data is not correctly stored or a file is missing, it returns an error.
+    ///
+    /// If the image is not stored, it does not attempt to pull the image from remote.
+    pub fn fetch_image_details(&mut self, image_name: &str) -> Result<ImageDetails> {
+        if self.config.is_none() {
+            self.config = Some(self.check_stored_image(image_name)?);
+        }
+        let image_config = self.config.as_ref().ok_or(EnclaveBuildError::ConfigError)?;
```
lauralt, line 423: I think you can have these in a single block by using `unwrap_or(_else?)`.

enclave_build/src/storage.rs
```diff
@@ -567,4 +595,29 @@ pub mod tests {
 
         assert_eq!(config, expected_config);
     }
+
+    #[test]
+    fn test_get_image_hash_from_name() {
```
lauralt, line 646: is the test name correct?

enclave_build/src/storage.rs
```diff
@@ -567,4 +595,29 @@ pub mod tests {
...
+        let (_root_path, mut storage) = setup_temp_storage();
+
+        let image_details = storage
+            .fetch_image_details(image::tests::TEST_IMAGE_NAME)
```
lauralt, line 650: Is self.config here None or Some? Can we test here the other scenario as well, e.g. if self.config is Some, the config is preserved?

raulmldv: I'm pretty sure that the config is `Some` after performing `setup_temp_storage` (does the store as well). Do we want to add a test with `None`?

lauralt: Ideally every code path should be tested, so if it's not too difficult, yes we should add a test for that.

enclave_build/src/storage.rs
lauralt: Ah right, if it's too complicated we can leave it as it is.

enclave_build/docs/storage.md
```diff
@@ -0,0 +1,171 @@
...
+|________oci-layout
+|            blobs/sha256/
+|            |_______98806831de537d56f19b4b3cbf6ed80187374a1521a18d2d0c5f8a0e3962b2ae -> manifest_hash points to config and layers
+|            |_______feb5d9fea6a5e9606aa995e879d862b825965ba48de054caab5ef356dc6b3412 -> config_hash
+|            |_______2db29710123e3e53a794f2694094b9b4338aa9ee5c40b930cb8063a1be392c54 -> layer_hash
+```
+
+### Examples
+
+To better understand the storage structure hand how the index and manifest point to other files, we
```
lauralt, line 35: hand -> and

enclave_build/docs/storage.md
```diff
@@ -0,0 +1,171 @@
...
+increases with each newly stored image. The manifest hash points us to the manifest blob.
+
+#### Layout file
+
+oci-layout:
+```
+{"imageLayoutVersion":"1.0.0"}
+```
+
+Simply defines the OCI image layout version we are using.
```
lauralt, line 67: missing subject

enclave_build/docs/storage.md
```diff
@@ -0,0 +1,171 @@
...
+manifest entry is added to the index and the blobs are written, each to their file.
+
+### Fetch details
+
+This operation returns the details of an image (image name, config digest and the config itself).
+These details are saved in the memory as the config of the image we are currently using to build
+the EIF. If a fetch was performed before, the config is already loaded and we can just return it.
+No other file read is needed.
+
+When we fetch an image without the config already loaded, we want to also that it is stored
```
lauralt, line 131: we want to also check?

enclave_build/docs/storage.md
```diff
@@ -0,0 +1,171 @@
+# Image Local Storage
```
lauralt, line 1: The doc looks much better now, thanks for the updates!

lauralt (dismissed):

about.toml
```diff
@@ -4,11 +4,12 @@ accepted = [
     "Zlib",
```
lauralt, line 4: the commit message needs some fixes (the signed offs are done twice).

enclave_build/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "enclave_build"
 version = "0.1.0"
 authors = ["The AWS Nitro Enclaves Team <aws-nitro-enclaves-devel@amazon.com>"]
 edition = "2018"
-rust-version = "1.58"
+rust-version = "1.59"
```
lauralt, line 6: Shouldn't we update this in all crates?

lauralt (approved):

PR #437, opened 2023-01-16
CI: use get-login-password instead of get-login

petreeftime:
It's not recommended to use get-login anymore, since the temporary password is printed out.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -121,6 +121,7 @@ do
 
 	configure_ne_driver
 
+	timeout 5m \
 	./build/nitro_cli/"${ARCH}"-unknown-linux-musl/release/deps/"${test_exec_name}" \
 		--test-threads=1 --nocapture || test_failed
 done < <(grep -v '^ *#' < build/test_executables.txt)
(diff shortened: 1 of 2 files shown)


run_tests.sh
```diff
@@ -14,7 +14,11 @@ export NITRO_CLI_BLOBS="${SCRIPTDIR}/blobs"
 export NITRO_CLI_ARTIFACTS="${SCRIPTDIR}/build"
 ARCH="$(uname -m)"
 
-$(aws ecr get-login --no-include-email --region us-east-1)
+AWS_ACCOUNT_ID=667861386598
```
meerd, line 17: nit: Could these variables be read from an external file and then exported?

alcioa (approved):

run_tests.sh
petreeftime: Potentially, yes. But what would be the benefit?

meerd: For better maintainability. We could combine (not only the variables above but) all hard-coded values in a file and source it from any of the scripts.

meerd: I also see some hard-coded docker registry addresses (667861386598.dkr.ecr.us-east-1.amazonaws.com) in this file that could be replaced by $ECR_URL.

petreeftime: Makes sense, but I'd like to do that in another patch series. I'd actually like to try and drop the private ECR containers as dependencies, so these tests can just be run by anyone from anywhere.

meerd (approved):

PR #438, opened 2023-01-17
enclave_build: Image manager trait

raulmldv:
Image Manager

The image manager trait exposes the operations required for building the EIF. The decision wether to fetch the image details from the local OCI storage or if they should be pulled from a registry is implemented as a failover logic that will allow the build to continue even if the image is missing locally, the storage is corrupt or missing.

The same manager trait will be used for building EIFs from Docker images but will be adapted in a future PR with the implied workflow changes in `lib.rs`.

### DIFF
--- a/enclave_build/src/image.rs
+++ b/enclave_build/src/image.rs
@@ -49,6 +49,10 @@ impl ImageDetails {
             config: deserialize_from_reader(image_data.config.data.as_slice())?,
         })
     }
+
+    pub fn config(&self) -> &ImageConfiguration {
+        &self.config
+    }
 }
 
 /// URIs that are missing a domain will be converted to a reference using the Docker defaults.
--- a/enclave_build/src/image.rs
+++ b/enclave_build/src/image.rs
@@ -4,6 +4,7 @@
 use crate::{EnclaveBuildError, Result};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::io::Read;
+use tempfile::NamedTempFile;
 
 use sha2::Digest;
 
@@ -52,6 +53,40 @@ impl ImageDetails {
 
     pub fn config(&self) -> &ImageConfiguration {
         &self.config
+    }
+
+    /// Extracts from the image and returns the CMD and ENV expressions (in this order).
+    ///
+    /// If there are no CMD expressions found, it tries to locate the ENTRYPOINT command.
+    /// The returned result will contain the files needed for the YAML generator.
+    pub fn extract_expressions(&self) -> Result<(NamedTempFile, NamedTempFile)> {
+        // Get the expressions from the image
+        let config_section = self
+            .config
+            .config()
+            .as_ref()
+            .ok_or(EnclaveBuildError::ConfigError)?;
+
+        let cmd = config_section.cmd();
+        let env = config_section.env();
+        let entrypoint = config_section.entrypoint();
+
+        // If no CMD instructions are found, try to locate an ENTRYPOINT command
+        let (cmd, env) = match (cmd, env, entrypoint) {
+            (Some(cmd), Some(env), _) => Ok((cmd.to_vec(), env.to_vec())),
+            (_, Some(env), Some(entrypoint)) => Ok((entrypoint.to_vec(), env.to_vec())),
+            (_, _, Some(entrypoint)) => Ok((entrypoint.to_vec(), Vec::new())),
+            (_, _, _) => Err(EnclaveBuildError::ExtractError(
+                "Failed to locate ENTRYPOINT".to_string(),
+            )),
+        }?;
+
+        let cmd_file = crate::docker::write_config(cmd)
+            .map_err(|err| EnclaveBuildError::ExtractError(format!("{:?}", err)))?;
+        let env_file = crate::docker::write_config(env)
+            .map_err(|err| EnclaveBuildError::ExtractError(format!("{:?}", err)))?;
+
+        Ok((cmd_file, env_file))
     }
 }
 
--- a/enclave_build/src/image_manager.rs
+++ b/enclave_build/src/image_manager.rs
@@ -4,11 +4,24 @@
 
 use std::convert::TryFrom;
 
+use log::warn;
 use oci_distribution::Reference;
+use tempfile::NamedTempFile;
 
 use crate::image::ImageDetails;
 use crate::storage::OciStorage;
 use crate::{EnclaveBuildError, Result};
+
+/// Trait which provides an interface for handling an image (OCI or Docker)
+pub trait ImageManager {
+    fn image_name(&self) -> &str;
+    /// Inspects the image and returns its metadata in the form of a JSON Value
+    fn inspect_image(&self) -> Result<serde_json::Value>;
+    /// Returns the architecture of the image
+    fn architecture(&self) -> Result<String>;
+    /// Returns two temp files containing the CMD and ENV expressions extracted from the image
+    fn extract_expressions(&self) -> Result<(NamedTempFile, NamedTempFile)>;
+}
 
 pub struct OciImageManager {
     /// Name of the container image.
@@ -25,16 +38,15 @@ impl OciImageManager {
         let image_name = normalize_tag(image_name)?;
 
         // The docker daemon is not used, so a local storage needs to be created
-        let storage =
-            match OciStorage::get_default_root_path().map_err(|err| eprintln!("{:?}", err)) {
-                Ok(root_path) => {
-                    // Try to create/read the storage. If the storage could not be created, log the error
-                    OciStorage::new(&root_path)
-                        .map_err(|err| eprintln!("{:?}", err))
-                        .ok()
-                }
-                Err(_) => None,
-            };
+        let storage = match OciStorage::get_default_root_path().map_err(|err| warn!("{:?}", err)) {
+            Ok(root_path) => {
+                // Try to create/read the storage. If the storage could not be created, log the error
+                OciStorage::new(&root_path)
+                    .map_err(|err| warn!("{:?}", err))
+                    .ok()
+            }
+            Err(_) => None,
+        };
 
         let image_details = Self::fetch_image_details(&image_name, storage).await?;
 
@@ -60,11 +72,7 @@ impl OciImageManager {
 
         let image_details = if let Some(storage) = local_storage {
             // Try to fetch the image from the storage
-            storage.fetch_image_details(image_name).map_err(|err| {
-                // Log the fetching error
-                eprintln!("{:?}", err);
-                err
-            })
+            storage.fetch_image_details(image_name)
         } else {
             Err(EnclaveBuildError::OciStorageNotFound(
                 "Local storage missing".to_string(),
@@ -74,7 +82,8 @@ impl OciImageManager {
         // If the fetching failed, pull it from remote and store it
         match image_details {
             Ok(details) => Ok(details),
-            Err(_) => {
+            Err(err) => {
+                warn!("Fetching from storage failed: {}", err);
                 // The image is not stored, so try to pull and then store it
                 let image_data = crate::pull::pull_image_data(image_name).await?;
 
@@ -82,7 +91,7 @@ impl OciImageManager {
                 if let Some(local_storage) = storage.as_mut() {
                     local_storage
                         .store_image_data(image_name, &image_data)
-                        .map_err(|err| eprintln!("Failed to store image: {:?}", err))
+                        .map_err(|err| warn!("Failed to store image: {:?}", err))
                         .ok();
                 }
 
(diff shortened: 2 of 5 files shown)


enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,144 @@
...
+        if let Some(storage) = local_storage {
+            // Try to fetch the image from the storage
+            let image_details = storage.fetch_image_details(&image_name).map_err(|err| {
+                // Log the fetching error
+                eprintln!("{:?}", err);
+                err
+            });
+
+            // If the fetching failed, pull it from remote and store it
+            return image_details.or(self.fetch_and_try_store_image(&image_name).await);
```
lauralt, line 94:
```rust
            return storage
                .fetch_image_details(&image_name)
                .map_err(|err| {
                    // Log the fetching error
                    eprintln!("{:?}", err);
                    err
                })
                .or(self.fetch_and_try_store_image(&image_name).await);
```

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        // Try to create/read the storage
+        let storage = match OciStorage::new(&root_path) {
+            Ok(manager) => Some(manager),
+            Err(err) => {
+                // If the storage could not be created, log the error
+                eprintln!("{:?}", err);
+                None
+            }
+        };
```
lauralt, line 64: I think we can use `map_err(...).ok()` here.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Returns a struct containing image metadata.
+    ///
+    /// If the image is stored correctly, the function tries to fetch the image from the storage.
+    ///
+    /// If the image is not stored or a storage was not created (the 'storage' field is None),
+    /// it pulls the image, stores it (if the 'storage' field is not None) and returns its metadata.
+    ///
+    /// If the pull succeeded but the store operation failed, it returns the pulled image metadata.
+    async fn get_image_details(&mut self, image_name: &str) -> Result<ImageDetails> {
```
lauralt, line 80: Why do we need a mutable reference? From the function name, I would imagine self is not mutated. I think the problem comes from the fact that we are doing more things than we should in a single function, maybe we can refactor a bit (do the storing separately). Is that doable?

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,144 @@
...
+                // Log the fetching error
+                eprintln!("{:?}", err);
+                err
+            });
+
+            // If the fetching failed, pull it from remote and store it
+            return image_details.or(self.fetch_and_try_store_image(&image_name).await);
+        }
+
+        self.fetch_and_try_store_image(&image_name).await
```
lauralt, line 97: Can we refactor this a bit so that we call `fetch_and_try_store_image` only once?

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        let cmd = config_section.cmd();
+        let env = config_section.env();
+        let entrypoint = config_section.entrypoint();
+
+        // If no CMD instructions are found, try to locate an ENTRYPOINT command
+        match (cmd, env, entrypoint) {
+            (Some(cmd), Some(env), _) => Ok((cmd.to_vec(), env.to_vec())),
+            (_, Some(env), Some(entrypoint)) => Ok((entrypoint.to_vec(), env.to_vec())),
+            (_, _, Some(entrypoint)) => Ok((entrypoint.to_vec(), Vec::<String>::new())),
```
lauralt, line 146: nit: no need for the explicit `String`.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,144 @@
...
+        // Get the image metadata from the pulled struct
+        let image_details = ImageDetails::build_details(image_name, &image_data)?;
+
+        Ok(image_details)
+    }
+
+    /// Extracts from the image and returns the CMD and ENV expressions (in this order).
+    ///
+    /// If there are no CMD expressions found, it tries to locate the ENTRYPOINT command.
+    fn extract_image(&mut self) -> Result<(Vec<String>, Vec<String>)> {
```
lauralt, line 122: Same comment here: the method should take an immutable reference judging by its name.

enclave_build/src/image_manager.rs
```diff
@@ -6,8 +6,25 @@ use crate::image::ImageDetails;
 use crate::storage::OciStorage;
 use crate::{EnclaveBuildError, Result};
 
+use tempfile::NamedTempFile;
```
lauralt, line 9: Can we use the one from vmm-sys-util?

enclave_build/src/image_manager.rs
```diff
@@ -6,8 +6,25 @@ use crate::image::ImageDetails;
...
 use tokio::runtime::Runtime;
 
+/// Trait which provides an interface for handling images.
+pub trait ImageManager {
+    fn image_name(&self) -> &String;
```
lauralt, line 14: `&String` is quite unusual, can we replace with `&str`?

enclave_build/src/image_manager.rs
```diff
@@ -223,3 +223,43 @@ impl ImageManager for OciImageManager {
         runtime.block_on(act_get_image)
     }
 }
+
+#[cfg(test)]
```
lauralt, line 227: The unit test shouldn't be in a separate commit than the code it is testing.

enclave_build/src/image_manager.rs
raulmldv: The idea of this image manager is to hide the storage implementation while allowing the EIF to be built even if we don't have the image stored. We have to mutate the storage at some point and I think it complicates things if we move it higher in the workflow.

enclave_build/src/image_manager.rs
raulmldv: For now we match the `load` operation from the Docker workflow. I suggest we change it everywhere in another PR and remove the `tempfile` dependency altogether. Right now I would have to modify too many modules that are not related to these changes.

lauralt: Got it, thanks

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,261 @@
...
+        // If the fetching failed, pull it from remote and store it
+        image_details.or(self.fetch_and_try_store_image(&image_name).await)
+    }
+
+    /// Pulls image from remote registry and stores it if possible
+    async fn fetch_and_try_store_image(&mut self, image_name: &str) -> Result<ImageDetails> {
+        // The image is not stored, so try to pull and then store it
+        let image_data = crate::pull::pull_image_data(image_name).await?;
+
+        // If the store operation failed, still return the image details
```
lauralt, line 101: This comment should be after attempting the store, it is confusing otherwise.

raulmldv:
I addressed the comments and redesigned the interface for the `ImageManager` trait. I applied suggested renaming and deleted some functions which imply modifying the Docker workflow. I decided to leave those changes to the following PR and focus on OCI for this one.

The trait is now simplified (removed build and pull operations) and no longer takes mutable references for the manager. To achieve that I decided to keep the image details at the image manager level (instead of keeping the storage). The details will be fetched on manager creation either from the local storage or pulled from the registry.

shtaked (approved):

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,251 @@
...
+// SPDX-License-Identifier: Apache-2.0
+#![allow(dead_code)]
+
+use tempfile::NamedTempFile;
+
+use crate::image::ImageDetails;
+use crate::storage::OciStorage;
+use crate::{EnclaveBuildError, Result};
+
+/// Trait which provides an interface for handling images.
```
belveryin, line 11: Nit: maybe this could be rephrased to "handling a particular image". The first time I read it I thought it would manage a collection of images.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,251 @@
...
+                "Local storage missing".to_string(),
+            ))
+        };
+
+        // If the fetching failed, pull it from remote and store it
+        match image_details {
+            Ok(details) => Ok(details),
+            Err(err) => {
+                // Log error
+                eprintln!("Fetching from storage failed: {}", err);
```
belveryin, line 88: Nit: if we fail in line 72, the error is logged twice: in line 74 and also here, in 88.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///
+    /// If the image is stored correctly, the function tries to fetch the image from the storage.
+    ///
+    /// If the image is not stored or a storage was not created (the 'storage' field is None),
+    /// it pulls the image, stores it (if the 'storage' field is not None) and returns its metadata.
+    ///
+    /// If the pull succeeded but the store operation failed, it returns the pulled image metadata.
+    async fn fetch_image_details(
+        image_name: &str,
+        mut storage: Option<OciStorage>,
```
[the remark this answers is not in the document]
lauralt: Also, why not have the parameter as `Option<&mut OciStorage>` so that you avoid doing the first instruction in the function?

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,178 @@
...
+                    local_storage
+                        .store_image_data(image_name, &image_data)
+                        .map_err(|err| eprintln!("Failed to store image: {:?}", err))
+                        .ok();
+                }
+
+                // Get the image metadata from the pulled struct
+                let image_details = ImageDetails::build_details(image_name, &image_data)?;
+
+                Ok(image_details)
```
lauralt, line 104: nit: -> `ImageDetails::build_details(image_name, &image_data)`

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    }
+}
+
+/// Adds the default ":latest" tag to an image if it is untagged
+fn normalize_tag(image_name: &str) -> String {
+    let name = image_name.to_string();
+    match name.contains(':') {
```
lauralt, line 113: Is any name that contains at least one ":" a valid tagged image? Is it valid to have "::" for example?

enclave_build/src/image_manager.rs
lauralt: I think we can keep the error logging only here.

enclave_build/src/image_manager.rs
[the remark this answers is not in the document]
raulmldv: As fetching from the storage and storing would both mutate the storage, later we need a binding instead of the first line.

enclave_build/src/image_manager.rs
raulmldv: Found a tag parser in `oci_distribution`

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        let image_details = if let Some(storage) = local_storage {
+            // Try to fetch the image from the storage
+            storage.fetch_image_details(image_name).map_err(|err| {
+                // Log the fetching error
+                eprintln!("{:?}", err);
+                err
+            })
+        } else {
+            Err(EnclaveBuildError::OciStorageNotFound(
```
lauralt, line 81: Do we ignore this error later? I'll try to refactor a bit this method.

shtaked (approved):

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,180 @@
...
+        image_name: &str,
+        mut storage: Option<OciStorage>,
+    ) -> Result<ImageDetails> {
+        let local_storage = storage.as_mut();
+
+        let image_details = if let Some(storage) = local_storage {
+            // Try to fetch the image from the storage
+            storage.fetch_image_details(image_name).map_err(|err| {
+                // Log the fetching error
+                eprintln!("{:?}", err);
```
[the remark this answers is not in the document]
raulmldv: I will change the logging.

enclave_build/src/image_manager.rs
raulmldv: We were logging it before in line 78 but decided to only log once. I think I misunderstood and removed that.

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,180 @@
...
+        } else {
+            Err(EnclaveBuildError::OciStorageNotFound(
+                "Local storage missing".to_string(),
+            ))
+        };
+
+        // If the fetching failed, pull it from remote and store it
+        match image_details {
+            Ok(details) => Ok(details),
+            Err(_) => {
```
petreeftime, line 89: I think this err might be catching a few too many things. Yes, it might be that the image is not already in storage, but it could also be that we don't have access to the storage (ie. permission issues) or that the folder doesn't exist and it might have to be created either by this library or, I think, preferably by the user.

enclave_build/src/image.rs
```diff
@@ -53,6 +54,40 @@ impl ImageDetails {
...
+
+        // If no CMD instructions are found, try to locate an ENTRYPOINT command
+        let (cmd, env) = match (cmd, env, entrypoint) {
+            (Some(cmd), Some(env), _) => Ok((cmd.to_vec(), env.to_vec())),
+            (_, Some(env), Some(entrypoint)) => Ok((entrypoint.to_vec(), env.to_vec())),
+            (_, _, Some(entrypoint)) => Ok((entrypoint.to_vec(), Vec::new())),
+            (_, _, _) => Err(EnclaveBuildError::ExtractError(
+                "Failed to locate ENTRYPOINT".to_string(),
+            )),
+        }?;
```
petreeftime, line 82: More readable as
```
let env = config_section.env().ok_or_default();
let cmd = config_section.cmd()
    .ok_or(|| config_section.entrypoint())
    .map_err(...)?;
```

enclave_build/src/image_manager.rs
```diff
@@ -0,0 +1,180 @@
...
+
+#[cfg(test)]
+pub mod tests {
+    use sha2::Digest;
+
+    use super::{normalize_tag, OciImageManager};
+
+    #[cfg(target_arch = "x86_64")]
+    const SAMPLE_IMAGE: &str =
+        "667861386598.dkr.ecr.us-east-1.amazonaws.com/enclaves-samples:vsock-sample-server-x86_64";
```
petreeftime, line 155: Let's publish some empty test images into our Public ECR so that anyone can pull these instead.

lauralt (approved):

shtaked (approved):

PR #508, opened 2023-05-24
Image Manager trait for Docker

raulmldv:
The build EIF process should have the same workflow for both OCI and Docker. In the previous PR #438 I introduced and implemented the Image Manager trait for OCI while changing the workflow to pull/build the required container image on manager instantiation. This way subsequent pulls are no longer needed for inspect operations.

The Docker workflow was using the docker daemon to access the local image storage (still using the pull API in Rust) thus having a different interface that the trait offers. With this refactoring we can manipulate both types of images with the same operations.

### DIFF
--- a/enclave_build/src/image_manager.rs
+++ b/enclave_build/src/image_manager.rs
@@ -7,6 +7,7 @@ use std::convert::TryFrom;
 use log::warn;
 use oci_distribution::Reference;
 use tempfile::NamedTempFile;
+use tokio::runtime::Runtime;
 
 use crate::image::ImageDetails;
 use crate::storage::OciStorage;
@@ -33,7 +34,7 @@ pub struct OciImageManager {
 impl OciImageManager {
     /// When calling this constructor, it also tries to initialize the storage at the default path.
     /// If this fails, the ImageManager is still created, but the 'storage' field is set to 'None'.
-    pub async fn new(image_name: &str) -> Result<Self> {
+    pub fn new(image_name: &str) -> Result<Self> {
         // Add the default ":latest" tag if the image tag is missing
         let image_name = normalize_tag(image_name)?;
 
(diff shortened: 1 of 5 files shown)


enclave_build/src/lib.rs
```diff
@@ -103,8 +102,14 @@ pub struct Docker2Eif<'a> {
 }
 
 impl<'a> Docker2Eif<'a> {
+    /// If the 'docker_dir' argument is Some (i.e. --docker-dir flag is used), the docker daemon will
+    /// always be used to build the image locally and store it in the docker cache.
+    ///
+    /// If the 'oci_image' argument is Some (i.e. the --image-uri flag is used) and the '--docker-dir'
```
shtaked, line 108: Not sure this part of comment is relevant here: there is no `oci_image` argument here. probably you wanted to write about `docker_image` argument and pulling the image with a help of docker daemon?

enclave_build/src/image_manager.rs
```diff
@@ -21,6 +22,8 @@ pub trait ImageManager {
     fn architecture(&self) -> Result<String>;
     /// Returns two temp files containing the CMD and ENV expressions extracted from the image
     fn extract_expressions(&self) -> Result<(NamedTempFile, NamedTempFile)>;
+    /// Returns true if the image manager uses the Docker daemon
+    fn use_docker(&self) -> bool;
```
shtaked, line 26: I find this method redundant, then the whole idea of making this trait abstract looses its sense.

enclave_build/src/lib.rs
```diff
@@ -266,43 +275,62 @@ impl<'a> Docker2Eif<'a> {
...
-                String::from_utf8_lossy(&output.stderr)
-            );
-            return Err(EnclaveBuildError::LinuxKitError(format!(
-                "{:?}",
-                String::from_utf8_lossy(&output.stderr)
-            )));
+        // If the docker daemon should be used, then call linuxkit with the '-docker' flag, which
+        // makes linuxkit search the image in the docker cache first.
+        // Otherwise, do not add the flag and let it pull the images itself, without docker.
+        if self.image_manager.use_docker() {
```
shtaked, line 287: At this moment we only initialize it as `DockerImageManager`. What's the point for this if here?
And if you plan to also initialize `OciImageManager` in this module then can not we store this `isDocker` externally rather providing a method from the inside?

shtaked (changes requested):

enclave_build/src/lib.rs
raulmldv: Left from older code, missed it.

enclave_build/src/lib.rs
raulmldv: Removing this function and replacing it with a variable or a type match.

Issue #553, opened 2023-11-01
vsock-proxy should lookup IPs of Host before accepting each connection


nv-sankalpk:
Currently, vsock-proxy gets the first domain name IP after lookup_host resolution and starts connecting to the same IP address. For connecting to services exposed through AWS ALB , AWS Loadbalancer IPs keeps changing, causing an exception of "Could not create connection". To resolve this , can we lookup the IPs of the domain on each client connection accept and then connect.

 let sockaddr = SocketAddr::new(self.remote_addr, self.remote_port);
        let sock_type = self.sock_type;
        self.pool.execute(move || {
            let mut server = match sock_type {
                SockType::Stream => TcpStream::connect(sockaddr)
                    .map_err(|_| format!("Could not connect to {:?}", sockaddr)),
                _ => Err("Socket type not implemented".to_string()),
            }
            .expect("Could not create connection");
            info!("Connected client from {:?} to {:?}", client_addr, sockaddr);

meerd:
Hi @nv-sankalpk,

The vsock-proxy was not originally designed with **ALB** as a use case in mind. However, your finding is valid and we can handle dynamically changing address for allow-listed hosts. For now, I prefer to keep the current lookup behavior as the default.

To accommodate your requirements, we could consider introducing a new command line argument, or a data field in the **vsock-proxy.yaml** file. This would allow enabling the additional behavior as needed.

mlschindler:
We have this issue as well.. we need vsock-proxy to be able to use ALBs and any other infra with potentially changing IP addresses.

meerd:
Hello @mlschindler & @nv-sankalpk,

The latest release of the nitro-cli (v1.3.0) resolves the issue you reported.

nv-sankalpk:
@meerd Trying to install the latest version :
amazon-linux-extras install aws-nitro-enclaves-cli=latest
..
Package aws-nitro-enclaves-cli-1.2.3-0.amzn2.x86_64 already installed and latest version
Is this fixed not yet released ?

PR #593, opened 2024-03-13
vsock_proxy: Perform DNS resolution after the expiration of the TTL

meerd:
**Description of changes:** This PR addresses the wrong assumption that domain names are resolved to fixed static IP addresses. With these changes, **vsock-proxy** periodically refreshes the resolved IP address of the remote host after the Time-To-Live (TTL) expires, ensuring that the most up-to-date IP address is used.

**Relevant issues:** #553 


vsock_proxy/Cargo.toml
```diff
@@ -12,6 +12,7 @@ chrono = "0.4"
 clap = "3.2"
 dns-lookup = "2.0.3"
```
foersleo, line 13: Looks like `dns-lookup` is not used anywhere anymore. We should probably remove it here as well.

vsock_proxy/src/dns.rs
```diff
@@ -18,35 +18,42 @@ pub fn resolve(addr: &str, ip_addr_type: IpAddrType) -> VsockProxyResult<Vec<Dns
 
     // DNS lookup
     // It results in a vector of IPs (V4 and V6)
-    let ips = lookup_host(&addr).map_err(|_| "DNS lookup failed!")?;
+    let resolver = Resolver::new(ResolverConfig::default(), ResolverOpts::default()).unwrap();
+
+    let rresults: Vec<DnsResolveResult> = resolver
+        .lookup_ip(addr)
+        .map_err(|error| format!("DNS lookup failed! {:?}", error))?
```
foersleo, line 25: This changes the format of the error message when the lookup fails. This makes `test_resolve_invalid_domain` unhappy as it matches on `DNS lookup failed!`. So the test probably needs an update to account for the extra info about the error type. I believe this is what makes the integration tests fail.

foersleo (commented):
Apart from the comments about the mismatch in error message and test pointed out before this looks good to me.

vsock_proxy/src/dns.rs
meerd: Nice catch! Fixed in the new revision (1415a7d).

foersleo (approved):

vsock_proxy/src/dns.rs
```diff
@@ -12,48 +12,50 @@ use std::net::IpAddr;
...
-        .map_err(|err| format!("Could not parse remote address: {}", err))?;
-
-    let ip = *addrs.first().ok_or("No IP address found")?;
-    let ttl = 60; // IMPORTANT TODO: Obtain this value dynamically
-
-    Ok(DnsResolveResult { ip, ttl })
+    let rresults = resolve(addr, ip_addr_type)?;
+    // Return the first resolved IP address and its TTL value.
+    // We already know that rresults containt at least one item.
+    Ok(*rresults.first().unwrap())
```
mariusknaust, line 67: We should be able to avoid the unwrap by `rresult.first().ok_or(...)`

vsock_proxy/src/dns.rs
```diff
@@ -18,35 +18,42 @@ pub fn resolve(addr: &str, ip_addr_type: IpAddrType) -> VsockProxyResult<Vec<Dns
 
     // DNS lookup
     // It results in a vector of IPs (V4 and V6)
-    let ips = lookup_host(&addr).map_err(|_| "DNS lookup failed!")?;
+    let resolver = Resolver::new(ResolverConfig::default(), ResolverOpts::default()).unwrap();
```
mariusknaust, line 21: We should use `map_err()` instead of unwrap here.

vsock_proxy/src/proxy.rs
```diff
@@ -64,11 +64,23 @@ pub fn check_allowlist(
...
-            let rresults = dns::resolve(addr, ip_addr_type)?;
-            for rresult in rresults.into_iter() {
-                if rresult.ip == remote_addr {
-                    info!("Matched with host IP \"{}\" and port \"{}\"", addr, port);
-                    return Ok(remote_addr);
+            let rresults = dns::resolve(addr, ip_addr_type);
+            let remote_addr_matched = rresults.clone().ok().and_then(|rresults| {
+                rresults
+                    .into_iter()
+                    .find(|rresult| rresult.ip == remote_addr)
```
mariusknaust, line 71: We should be able to simplify this by: `rresult.iter().flatten().find(...)`

mariusknaust (approved):

PR #595, opened 2024-03-19
Replace shiplift with bollard for Docker 25+ compatibility

meerd:
The `shiplift` crate is outdated and incompatible with Docker 25. This PR replaces the it with `bollard`, allowing `nitro-cli` to work with Docker version 25 and newer.

Special thanks to @jalaziz for their contribution.

### DIFF
--- a/enclave_build/src/docker.rs
+++ b/enclave_build/src/docker.rs
@@ -2,14 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::docker::DockerError::CredentialsError;
+use crate::utils::handle_stream_output;
 use base64::{engine::general_purpose, Engine as _};
 use bollard::auth::DockerCredentials;
 use bollard::image::{BuildImageOptions, CreateImageOptions};
 use bollard::secret::ImageInspect;
 use bollard::Docker;
 use flate2::{write::GzEncoder, Compression};
-use futures::stream::StreamExt;
-use log::{debug, error, info};
+use log::{debug, error};
 use serde_json::{json, Value};
 use std::fs::File;
 use std::io::Write;
@@ -196,30 +196,11 @@ impl DockerUtil {
                 }
             };
 
-            let mut stream =
-                self.docker
-                    .create_image(Some(create_image_options), None, credentials);
+            let stream = self
+                .docker
+                .create_image(Some(create_image_options), None, credentials);
 
-            loop {
-                if let Some(item) = stream.next().await {
-                    match item {
-                        Ok(output) => {
-                            if let Some(err_msg) = &output.error {
-                                error!("{:?}", err_msg);
-                                break Err(DockerError::PullError);
-                            } else {
-                                info!("{:?}", output);
-                            }
-                        }
-                        Err(e) => {
-                            error!("{:?}", e);
-                            break Err(DockerError::PullError);
-                        }
-                    }
-                } else {
-                    break Ok(());
-                }
-            }
+            handle_stream_output(stream, DockerError::PullError).await
         })
     }
 
@@ -244,7 +225,7 @@ impl DockerUtil {
         let runtime = Runtime::new().map_err(|_| DockerError::RuntimeError)?;
 
         runtime.block_on(async move {
-            let mut stream = self.docker.build_image(
+            let stream = self.docker.build_image(
                 BuildImageOptions {
                     dockerfile: "Dockerfile".to_string(),
                     t: self.docker_image.clone(),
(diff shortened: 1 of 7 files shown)


meerd:
Update: Testing is in progress. This PR will be merged soon.

foersleo (approved):
I have done some testing around this to see if this works as expected on current and older versions of docker. I used different base AMIs to get some older versions of docker to run. The results are summarized in the table below.

| OS | Docker version | Working without PR | Working with PR |
| - | - | - | - |
| Ubuntu 18 | 20.10.21 | :white_check_mark: | :white_check_mark: |
| Ubuntu 22 | 24.0.5 | :white_check_mark: | :white_check_mark: |
| AL2 | 17.12.1[^1] | :white_check_mark: | :white_check_mark: |
| AL2 | 20.10.25 | :white_check_mark: | :white_check_mark: |
| AL2023 | 25.0.3 | :x: | :white_check_mark: |

[^1]: For the testcase with Docker 17 on AL2, I had to downgrade the docker package and some other conflicting packages, so this test case is somewhat flaky. But it shows that we are backwards compatible until then.

I did not spend the extra time to go even further back in Docker version history, given that Docker 17.12 moved into EOL in February 2018.

Overall, I think the results show that the update works will with older Docker versions, while also enabling the use of nitro-cli with newer versions.

enclave_build/src/docker.rs
```diff
@@ -38,20 +41,25 @@ pub struct DockerUtil {
...
-        DockerUtil {
-            // DOCKER_HOST environment variable is parsed inside
-            // if docker daemon address needs to be substituted.
-            // By default it tries to connect to 'unix:///var/run/docker.sock'
-            docker: Docker::new(),
+        // DOCKER_HOST environment variable is parsed inside
+        // if docker daemon address needs to be substituted.
+        // By default, it tries to connect to 'unix:///var/run/docker.sock'
+        let docker = Docker::connect_with_defaults().map_err(|e| {
+            error!("{:?}", e);
```
mariusknaust, line 56: nit: Should be added to the `ConnectionError` itself and printed when it is handled

mariusknaust, line 56: nit: The error should be encapsulated into the `ConnectionError` and printed when it is handled.

enclave_build/src/docker.rs
```diff
@@ -160,46 +169,40 @@ impl DockerUtil {
...
+            let credentials = match self.get_credentials() {
+                Ok(auth) => Some(auth),
                 // It is not mandatory to have the credentials set, but this is
                 // the most likely reason for failure when pulling, so log the
                 // error.
                 Err(err) => {
                     debug!("WARNING!! Credential could not be set {:?}", err);
+                    None
                 }
             };
```
mariusknaust, line 197: nit: Looks like this is replicating existing methods of `Result`:
```
let credentials = self.get_credentials()
	.inspect_err(|err| debug!("WARNING!! Credential could not be set {:?}", err))
	.ok();
```

enclave_build/src/docker.rs
```diff
@@ -325,10 +346,15 @@ impl DockerUtil {
             }
 
             let act = async {
-                match self.docker.images().get(&self.docker_image).inspect().await {
+                match self.docker.inspect_image(&self.docker_image).await {
                     Ok(image) => Ok((
-                        image.config.entrypoint.unwrap(),
-                        image.config.env.ok_or_else(Vec::<String>::new).unwrap(),
+                        image.config.clone().unwrap().entrypoint.unwrap(),
```
mariusknaust: nit: Should have a comment why `unwrap` is safe, e.g. because it was checked before. Or even better would be to restructure the code to use the checked value from above.

enclave_build/src/docker.rs
```diff
@@ -272,118 +273,49 @@ impl DockerUtil {
...
+        let runtime = Runtime::new().map_err(|_| DockerError::RuntimeError)?;
+        let image_future = self.docker.inspect_image(&self.docker_image);
+
+        runtime.block_on(async {
+            match image_future.await {
+                Ok(image) => Ok(image),
                 Err(e) => {
                     error!("{:?}", e);
                     Err(DockerError::InspectError)
                 }
```
mariusknaust, line 252: nit: This could be handled by `map_err`.

enclave_build/src/docker.rs
```diff
@@ -63,67 +63,70 @@ impl DockerUtil {
...
+    /// Returns the credentials by reading ${HOME}/.docker/config.json or ${DOCKER_CONFIG}
+    ///
+    /// config.json doesn't seem to have a schema that we could use to validate
+    /// we are parsing it correctly, so the parsing mechanism had been infered by
+    /// reading a config.json created by:
+    //         Docker version 19.03.2
+    fn get_credentials(&self) -> Result<DockerCredentials, DockerError> {
+        let host = match Self::parse_docker_host(&self.docker_image) {
+            Some(host) => host,
+            None => return Err(CredentialsError("Invalid docker image URI!".to_string())),
```
mariusknaust, line 89: nit: This match could be handled using `.ok_or_else(|| CredentialsError(...))?`

mariusknaust (approved):

PR #650, opened 2025-01-20
Allow signing EIFs with KMS while building

atanzu:
This commit updates `nitro-cli` to use `aws-nitro-enclaves-image-format` v0.4 which allows to sign EIF during build using a key from KMS. In order to do so, user needs to specify KMS key ARN and region instead of the local private key file.

*Issue #, if available:*

*Description of changes:*


atanzu:
License check fails due to changes in `ring` crate: https://github.com/EmbarkStudios/cargo-about/issues/271

Security check fails due to https://github.com/rustsec/rustsec/issues/1318

src/common/commands_parser.rs
```diff
@@ -121,16 +125,18 @@ impl BuildEnclavesArgs {
     pub fn new_with(args: &ArgMatches) -> NitroCliResult<Self> {
         let signing_certificate = parse_signing_certificate(args);
         let private_key = parse_private_key(args);
+        let kms_key_id = parse_kms_key_id(args);
+        let kms_key_region = parse_kms_key_region(args);
 
-        match (&signing_certificate, &private_key) {
-            (Some(_), None) => {
+        match (&signing_certificate, &private_key, &kms_key_id) {
```
eugkoira, line 131: a bit redundant after we setup clap to verify all the rules with `group` and `require`

atanzu: So shall I just panic then?

eugkoira: otherwise if we ignore clap rules, this misses `(Some(_), None, None)` case, also `kms_key_region` is not involved here

tests/tests.rs
eugkoira, line 1: Would be great to have a full end-to-end test for the new functionality here

src/lib.rs
```diff
@@ -714,6 +720,23 @@ macro_rules! create_app {
                             .long("private-key")
                             .help("Local path to developer's Eliptic Curve private key."),
                     )
+                    .arg(
```
eugkoira, line 723: We should add tests for those group/require/multiple logic to `test_nitro_cli_args.rs`

tests/tests.rs
atanzu: Currently existing tests use signing with a local private key. We might want to add tests with KMS keys similar to ones in image-format crate, but those won't be executed in CI anyway, unless we extend the CI resources.

src/common/commands_parser.rs
eugkoira: may be use `assert`?

tests/tests.rs
eugkoira: I think we could extend our CI resources for such testing.

src/lib.rs
atanzu: Added new test cases in the follow-up commit.

eugkoira: Thanks!

tests/tests.rs
atanzu: Then it will be a separate PR to keep it more focused on the PR deliverable.

src/common/commands_parser.rs
mariusknaust: Since this function is meant to be called on the `clap::ArgMatches` and the options are just check (without unpacking them), we are just validating a precondition and an `assert` should be enough. But I don't have a strong opinion on this.

mariusknaust (approved):

src/common/commands_parser.rs
atanzu: I will remove this check completely because later on we still have another check (when creating a `SignKeyInfo`.

eugkoira (approved):

mariusknaust (approved):

PR #639, opened 2024-08-02
Support for PCR0 signing

zhanjh:
*Issue #, if available:*

*Description of changes:*

This PR is based on [aws-nitro-enclaves-image-format/pull/30](https://github.com/aws/aws-nitro-enclaves-image-format/pull/30)

- Add subcommand `measure-enclave` to generate measurements without writing output enclave image file
- Add subcommand `sign-pcr0` to sign PCR0 with private key and write output signature file
- Change subcommand `build-enclave` to accept signed signature.

For security reasons, the private key is usually in a separated environment, so it may not be possible to build an enclave with private key in a single step from the building pipeline. Inspired by [Digest Signing in oesign](https://github.com/openenclave/openenclave/blob/2517ec75c25b07fa6fdff815bac361f403f63856/docs/DesignDocs/oesign_digest_signing_support.md),

1. In building environment, generate PCR0 of the enclave
2. Send PCR0 to the environment containing the private key for signing
3. Back in the building environment, build enclave with certificate and signed signature. 

```shell
# 1. generate PCR0
nitro-cli measure-enclave --docker-uri example:latest

# 2. sign PCR0 in another environment with private key
nitro-cli sign-pcr0 \
  --pcr0 "pcr0-hash-in-hex" \
  --private-key key.pem \
  --output-file pcr0.sig


# 3. build enclave with certificate and signature
nitro-cli build-enclave \
  --docker-uri example:latest \
  --signing-certificate certificate.pem \
  --signature pcr0.sig \
  --output-file example.signed.eif
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #652, opened 2025-01-23
Sign EIFs separately

atanzu:
*Issue #, if available:*
#204 

*Description of changes:*

This PR allows to sign pre-built EIFs using the new `sign-eif` command. Users can use either a local private key, or a KMS key.

### DIFF
--- a/src/common/commands_parser.rs
+++ b/src/common/commands_parser.rs
@@ -259,6 +259,32 @@ impl PcrArgs {
     }
 }
 
+/// The arguments used by `sign-eif` command
+#[derive(Debug, Clone)]
+pub struct SignEifArgs {
+    /// Path to the EIF file needed for signing
+    pub eif_path: String,
+    /// The path to the signing certificate for signed enclaves.
+    pub signing_certificate: Option<String>,
+    /// ARN of the KMS key or path to the local private key for signed enclaves.
+    pub private_key: Option<String>,
+}
+
+impl SignEifArgs {
+    /// Construct a new `SignEifArgs` instance from the given command-line arguments.
+    pub fn new_with(args: &ArgMatches) -> NitroCliResult<Self> {
+        let signing_certificate = parse_signing_certificate(args);
+        let private_key = parse_private_key(args);
+
+        Ok(SignEifArgs {
+            eif_path: parse_eif_path(args)
+                .map_err(|e| e.add_subaction("Parse EIF path".to_string()))?,
+            signing_certificate,
+            private_key,
+        })
+    }
+}
+
 /// Parse file path to hash from the command-line arguments.
 fn parse_file_path(args: &ArgMatches, val_name: &str) -> NitroCliResult<String> {
     let path = args.get_one::<String>(val_name).ok_or_else(|| {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,8 @@ pub mod utils;
 
 use aws_nitro_enclaves_image_format::defs::eif_hasher::EifHasher;
 use aws_nitro_enclaves_image_format::utils::eif_reader::EifReader;
+use aws_nitro_enclaves_image_format::utils::eif_signer::EifSigner;
+use aws_nitro_enclaves_image_format::utils::SignKeyData;
 use aws_nitro_enclaves_image_format::{generate_build_info, utils::get_pcrs};
 use log::{debug, info};
 use sha2::{Digest, Sha384};
@@ -25,9 +27,9 @@ use std::convert::TryFrom;
 use std::fs::{File, OpenOptions};
 use std::io::{self, Read, Write};
 use std::os::unix::net::UnixStream;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
-use common::commands_parser::{BuildEnclavesArgs, EmptyArgs, RunEnclavesArgs};
+use common::commands_parser::{BuildEnclavesArgs, EmptyArgs, RunEnclavesArgs, SignEifArgs};
 use common::json_output::{
     EifDescribeInfo, EnclaveBuildInfo, EnclaveTerminateInfo, MetadataDescribeInfo,
 };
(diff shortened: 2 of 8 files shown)


src/lib.rs
```diff
@@ -824,5 +914,41 @@ macro_rules! create_app {
...
+            .subcommand(
+                Command::new("sign-eif")
+                    .about("Sign EIF with the given key")
+                    .arg(
+                        Arg::new("eif-path")
+                            .long("eif-path")
+                            .help("Path pointing to a prebuilt Eif image")
+                    )
+                    .arg(
+                        Arg::new("signing-certificate")
```
meerd, line 926: If we decide to retain the local signing option, this should be --signing-certificate-path for clarity.

src/lib.rs
```diff
@@ -824,5 +914,41 @@ macro_rules! create_app {
...
+                            .long("signing-certificate")
+                            .help("Local path to developer's X509 signing certificate."),
+                    )
+                    .arg(
+                        Arg::new("private-key")
+                            .long("private-key")
```
meerd, line 932: If we decide to retain the local signing option, this should be --private-key-path for clarity.

src/lib.rs
```diff
@@ -824,5 +914,41 @@ macro_rules! create_app {
...
+                            .long("eif-path")
+                            .help("Path pointing to a prebuilt Eif image")
+                    )
+                    .arg(
+                        Arg::new("signing-certificate")
+                            .long("signing-certificate")
+                            .help("Local path to developer's X509 signing certificate."),
+                    )
+                    .arg(
+                        Arg::new("private-key")
```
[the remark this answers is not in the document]
atanzu: My vision is, we might have customers who build EIFs in one (insecure) environment, and then sign those EIFs separately in a secure environment with a local private key, similar to #639.

src/lib.rs
atanzu: How I see it, the parameter names shall be the same as the ones used when signing an image during build. Updating syntax for `build-enclave` is a breaking change unless we keep both the old parameter name and the new one, plus add a deprecation warning.

src/common/commands_parser.rs
```diff
@@ -265,6 +265,62 @@ impl PcrArgs {
...
+
+impl SignEifArgs {
+    /// Construct a new `SignEifArgs` instance from the given command-line arguments.
+    pub fn new_with(args: &ArgMatches) -> NitroCliResult<Self> {
+        let signing_certificate = parse_signing_certificate(args);
+        let private_key = parse_private_key(args);
+        let kms_key_id = parse_kms_key_id(args);
+        let kms_key_region = parse_kms_key_region(args);
+
+        validate_signing_credentials(&signing_certificate, &private_key, &kms_key_id)?;
```
eugkoira, line 291: a bit weird that `kms_key_region` is not validated

src/lib.rs
eugkoira: I vote for consistency with what we already have

src/lib.rs
eugkoira: it will be inconsistent with `build-eif` experience and can bring more confusion then benefits

eugkoira (approved):

src/common/commands_parser.rs
eugkoira: no, I mean potentially we could have region and private key at the same time

src/lib.rs
```diff
@@ -824,5 +914,41 @@ macro_rules! create_app {
...
+                    .arg(
+                        Arg::new("kms-key-id")
+                            .long("kms-key-id")
+                            .help("Specify unique id of the KMS key")
+                    )
+                    .arg(
+                        Arg::new("kms-key-region")
+                            .long("kms-key-region")
+                            .help("Specify region in which the KMS key resides")
+                            .requires("kms-key-id")
```
eugkoira, line 944: also testing those rules same as for `build-eif` would be beneficial

src/lib.rs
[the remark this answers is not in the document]
meerd: We all know what could happen if a private key is leaked. :) Given the security risks associated with private key exposure, the users should be encouraged to use KMS as the recommended best practice. (Or using PKCS#11 in the future. This was also mentioned by @dwmw2 [here](https://github.com/aws/aws-nitro-enclaves-cli/pull/406#discussion_r1784343427).)

I believe there are two options:
- Omit local `private-key` signing for `sign-eif` command and mark the argument in `build-eif` as deprecated
- Keep `private-key` as an option for `sign-eif` but add warnings about security implications.

src/lib.rs
meerd: >we keep both the old parameter name and the new one, plus add a deprecation warning

Yes, this is what I meant. To clarify, `--signing-certificate` or `--private-key` should remain in place until the next major release of `nitro-cli`.

src/lib.rs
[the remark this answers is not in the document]
eugkoira: Customers can still have a secure isolated  environment for signing with a private key like in https://github.com/aws/aws-nitro-enclaves-cli/pull/639. Why should we restrict their possibilities?

We can specify in our documentation the recommended way but I don't understand the urge to remove it completely.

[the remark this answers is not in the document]
agraf: Customers can literally use Nitro Enclaves to receive the signing key from KMS and pass it as parameter here, right? I don't believe the tool should per se assume an insecure environment. To clarify environmental requirements and establish best practices is the task of our documentation.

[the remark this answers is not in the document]
dwmw2: Signing requires a private key. That's kind of fundamental. The question is *where* that private key is found. 

The real problem here seems to be that we have two completely separate command line arguments to provide the private key. There's the `--private-key` argument, which makes sense. A user will reasonably expect to be able to pass that a filename, an RFC7512 PKCS#11 URI, and maybe other kinds of things like a [TPM URI](https://datatracker.ietf.org/doc/html/draft-mavrogiannopoulos-tpmuri-01) referencing keys in non-volatile storage (although that's kind of dead now, and mostly a user would expect to provide a PEM file with a TPM-wrapped key and have that Just Work too.

And then we have the `--kms-key-id` argument. Which probably should have just accepted an ARN for a KMS key, e.g. `--private-key arn:aws:kms:us-west-2:111122223333:key/0987dcba-09fe-87dc-65ba-ab0987654321`, from the start? If we're talking about cleaning up the user interface, I'd much prefer us to deprecate `--kms-key-id` and use *only* `--private-key` to tell the tooling which private key to use.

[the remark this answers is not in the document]
atanzu: I agree that having a single parameter which dispatches internally basing on the provided value is the best option in terms of UX, and good thing that `--kms-key-id` has only been introduced to mainline, there is no actual release which fixed that in stone.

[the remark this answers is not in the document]
dwmw2: > good thing that `--kms-key-id` has only been introduced to mainline, there is no actual release which fixed that in stone.

Excellent. So let's fix that to use `--private-key arn:aws:kms:…` before ever releasing it.

src/common/commands_parser.rs
dwmw2: Which is why you could just pass the ARN of the KMS key in as the `--private-key` argument... :)

SPECS/aws-nitro-enclaves-cli.spec
```diff
@@ -195,6 +195,12 @@ fi
 %{ne_include_dir}/*
 
 %changelog
+* Thu Jan 23 2025 Mark Kirichenko <mkirich@amazon.de> - 1.4.0-0
```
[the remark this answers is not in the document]
atanzu: Ack, removed this commit from the PR branch

eugkoira (approved):

meerd (approved):

PR #749, opened 2026-05-12
Fix Clippy unnecessary_sort_by warning

mariusknaust:
Replace sort_by with sort_by_key(Reverse(...)) as suggested by Clippy. This lint is not auto-fixable because it is marked MaybeIncorrect.


mariusknaust:
Already covered by PR #748

PR #755, opened 2026-06-26
Add support for building enclaves from Amazon Linux kernels

jxwise:
*Issue:*
The CLI ships enclaves with a bundled static kernel blob; this extends the capability to build enclaves directly from Amazon Linux kernels.

*Description of changes:*
Makes the Amazon Linux kernel the default for `build-enclave`, fetched on demand from the AL2023 repositories.

Flags introduced to `build-enclave`:
- `--kernel-version`: build from a specific AL2023 kernel version; defaulting to the latest.
  - `list-packages` and `download-kernel` as sub-commands to inspect and fetch AL2023 kernel RPM's directly.
- `--kernel-rpm-path`: build from a local kernel RPM.
- `--legacy-kernel`: opt back into the bundled static blob; deprecated, with removal in v3.0.0.

Provide unit and integration test coverage across the extraction and AL build path.

Simultaneously prepare the release, requiring a major bump to 2.0.0 due to breaking changes; including fixes to the CI workflow.


jxwise:
> @[jxwise](https://github.com/jxwise) jxwise [force-pushed](https://github.com/aws/aws-nitro-enclaves-cli/compare/2f04740b0bd59cdc79eb33da8281af64db97ef23..6cfc56e91b135ad0832182cda6fc5f262f17cc17) the al-kernel branch from [2f04740](https://github.com/aws/aws-nitro-enclaves-cli/commit/2f04740b0bd59cdc79eb33da8281af64db97ef23) to [6cfc56e](https://github.com/aws/aws-nitro-enclaves-cli/commit/6cfc56e91b135ad0832182cda6fc5f262f17cc17)

Added `Co-authored-by: Leonard Foerster <foersleo@amazon.com>` to commit messages.

enclave_build/src/yaml_generator.rs
```diff
@@ -57,39 +78,174 @@ impl YamlGenerator {
...
             init_path,
-            nsm_path,
+            modules,
             cmd_path,
             env_path,
         }
     }
 
+    /// Compute the topological sort of modules based on dependencies.
+    /// Returns modules in load order (dependencies first, nsm.ko last).
```
amateurcow, line 95: Why does nsm.ko need to always be last? Could we document this?

enclave_build/src/yaml_generator.rs
```diff
@@ -57,39 +78,174 @@ impl YamlGenerator {
...
-                },
-                FileTemplate {
-                    path: String::from("nsm.ko"),
-                    source: self.nsm_path.clone(),
-                    mode: String::from("0755"),
-                },
-            ),
-        };
+        // Generate the module load order file
+        let load_order_content = self.generate_modules_load_order()?;
```
amateurcow, line 207: This also calls `compute_module_load_order`, do we need to call it twice? Or could we just call it once and generate the `load_order_content` here and re-use the result of compute_module_load_order later.

enclave_build/src/yaml_generator.rs
jxwise: There's no strict requirement that it is last; only that it loads after its dependencies.

Since nothing else depends on `nsm.ko`, it most naturally falls to be loaded last.

enclave_build/src/yaml_generator.rs
```diff
@@ -57,39 +78,174 @@ impl YamlGenerator {
...
+
+            result.push(module);
+
+            // Reduce in-degree of dependents
+            for dependent in dependents.get(module_name).unwrap() {
+                let degree = in_degree.get_mut(dependent).unwrap();
+                *degree -= 1;
+                if *degree == 0 && *dependent != "nsm" {
+                    queue.push(dependent);
+                    queue.sort(); // Keep sorted for deterministic output
```
amateurcow, line 157: nit: Can we insert sorted instead of sorting in every iteration of the loop? Could blow up pretty fast if there are a large number of dependencies, although unlikely to be enough to be impactful.

blobs/aarch64/init
amateurcow, line 1: Were these blobs built from this PR - https://github.com/aws/aws-nitro-enclaves-sdk-bootstrap/pull/52?

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+    pub kernel_config: PathBuf,
+    /// List of extracted modules with their dependencies
+    pub modules: Vec<ModuleInfo>,
+    /// Path to modules manifest file
+    pub modules_manifest: PathBuf,
+}
+
+/// Information about a kernel module
+#[derive(Debug, Clone)]
+pub struct ModuleInfo {
```
amateurcow, line 60: Does this match with ModuleEntry definition from previous commit? Might be better to have a common def for future proofing.

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+        Ok(output) if output.status.success() => {
+            let deps_str = String::from_utf8_lossy(&output.stdout);
+            deps_str
+                .trim()
+                .split(',')
+                .filter(|s| !s.is_empty())
+                .map(|s| s.trim().to_string())
+                .collect()
+        }
+        _ => vec![],
```
amateurcow, line 136: Should we log a warning if `modinfo` failed?

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+        let mode = entry.entry().mode();
+        let name = entry.entry().name().to_string();
+        let rel = Path::new(&name);
+        let rel = rel
+            .strip_prefix("./")
+            .or_else(|_| rel.strip_prefix("/"))
+            .unwrap_or(rel);
+        let out_path = dest.join(rel);
+
+        let is_regular = mode & 0o170000 == 0o100000;
```
amateurcow, line 250: Magic numbers?

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+        .map(|e| e.path())
+        .ok_or_else(|| {
+            new_nitro_cli_failure!(
+                "No kernel version directory found",
+                NitroCliErrorEnum::FileOperationFailure
+            )
+        })?;
+
+    // Kernel image is typically in vmlinuz or vmlinux
+    let image_name = match arch {
```
amateurcow, line 298: All of the options provided here were just vmlinuz, what is the purpose of the aarch check here at present? Is there a different name that should be used on aarch64?

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+    };
+
+    let image_path = version_dir.join(image_name);
+    if image_path.exists() {
+        return Ok(image_path);
+    }
+
+    // Try alternate locations
+    let boot_vmlinuz = extract_dir.join("boot").join(format!(
+        "vmlinuz-{}",
```
amateurcow, line 311: Likewise, should this be searching for `image_name` as well? Or are we always searching for `vmlinux`

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+        "Kernel config not found",
+        NitroCliErrorEnum::FileOperationFailure
+    ))
+}
+
+/// Find modules directory in extracted RPM
+fn find_modules_dir(extract_dir: &Path) -> NitroCliResult<PathBuf> {
+    let modules_base = extract_dir.join("lib/modules");
+
+    // Find the kernel version directory
```
amateurcow, line 370: This `version_dir` finding logic is pretty much duplicated across all 3 of these functions, could we abstract this out?

src/kernel_extract.rs
```diff
@@ -0,0 +1,979 @@
...
+    Ok(KernelExtractInfo {
+        kernel_image: kernel_image_dest,
+        kernel_config: kernel_config_dest,
+        modules: extracted_modules,
+        modules_manifest: manifest_path,
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
amateurcow, line 836: nit: Tests seem to be arm64 targeted, should we have a couple of tests for other functionality such as config parsing/rpm extraction?

src/main.rs
amateurcow, line 1: nit: copyright update

src/rpm_repo.rs
```diff
@@ -0,0 +1,364 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
amateurcow, line 1: nit: copyright date

src/main.rs
```diff
@@ -316,6 +318,42 @@ fn main() {
...
+            let list_args = ListPackagesArgs::new_with(args);
+            let packages = rpm_repo::list_kernel_packages(&list_args.arch)
+                .map_err(|e| {
+                    e.add_subaction("Failed to list packages".to_string())
+                        .set_action(LIST_PACKAGES_STR.to_string())
+                })
+                .ok_or_exit_with_errno(None);
+
+            if packages.is_empty() {
+                println!("No kernel packages found with version >= 6.12");
```
amateurcow, line 331: Is there a reason why this is restricted to kernel >= v6.12?

src/rpm_repo.rs
```diff
@@ -0,0 +1,364 @@
...
+use crate::new_nitro_cli_failure;
+
+/// Mirror list URLs for Amazon Linux 2023
+const AL2023_MIRROR_AARCH64: &str =
+    "https://cdn.amazonlinux.com/al2023/core/mirrors/latest/aarch64/mirror.list";
+const AL2023_MIRROR_X86_64: &str =
+    "https://cdn.amazonlinux.com/al2023/core/mirrors/latest/x86_64/mirror.list";
+
+/// Minimum kernel version to list (6.12 and later)
+const MIN_KERNEL_VERSION: &str = "6.12.0";
```
amateurcow, line 23: Same as above, why is the minimum version shown v6.12?

src/rpm_repo.rs
```diff
@@ -0,0 +1,364 @@
...
+        .filter(|pkg| {
+            is_main_kernel_package(&pkg.name)
+                && is_version_at_least(&pkg.version.version, MIN_KERNEL_VERSION)
+        })
+        .map(package_to_info)
+        .collect();
+
+    // Sort by name and version
+    filtered.sort_by(|a, b| match a.name.cmp(&b.name) {
+        std::cmp::Ordering::Equal => a.version.cmp(&b.version),
```
amateurcow, line 174: This seems to be lexicographical comparison since version is a string (so it would be sorted as v6.1, v6.11, v6.12, ... , v6.2, v6.3)

src/lib.rs
```diff
@@ -78,9 +86,130 @@ pub fn build_from_docker(
...
+    legacy_kernel: bool,
+    kernel_version: &Option<String>,
+    kernel_rpm_path: &Option<String>,
 ) -> NitroCliResult<(File, BTreeMap<String, String>)> {
     let blobs_path =
         blobs_path().map_err(|e| e.add_subaction("Failed to retrieve blobs path".to_string()))?;
+
+    // Determine architecture
+    let arch = match std::env::consts::ARCH {
+        "aarch64" => "aarch64",
```
amateurcow, line 98: nit: Could we just `let arch = std::env::consts::ARCH` and then have a check to return the error if needed?

src/lib.rs
```diff
@@ -796,6 +910,25 @@ macro_rules! create_app {
...
+                            .long("legacy-kernel")
+                            .help("[DEPRECATED] Build with the bundled static kernel blob instead of the Amazon Linux kernel. This flag will be removed in the next major version (3.0.0).")
+                            .action(clap::ArgAction::SetTrue)
+                            .conflicts_with("kernel-version")
+                            .conflicts_with("kernel-rpm-path"),
+                    )
+                    .arg(
+                        Arg::new("kernel-version")
+                            .long("kernel-version")
+                            .help("Specific Amazon Linux kernel version to use (defaults to latest)"),
```
amateurcow, line 925: Is it worth noting the kernel version restriction imposed by the previous patch? Assuming that restriction is kept.

tests/integration/kernel_helpers.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
amateurcow, line 1: nit: copyright

tests/tests.rs
```diff
@@ -84,6 +84,9 @@ mod tests {
             img_name: None,
```
amateurcow, line 84: Is the intention of these changes just to make the existing tests run? If so, would it make sense to run with different combinations of args here to exercise the other paths as well for all the existing testing coverage?

.cargo/audit.toml
```diff
@@ -0,0 +1,27 @@
+# cargo-audit configuration.
```
amateurcow, line 1: Might be worth including in the commit message that the default behaviour is changing (i.e. `--legacy-kernel` will be needed for the existing behaviour)

Issue #756, opened 2026-07-01
`build-enclave` with a cross-account KMS key silently produces an UNSIGNED EIF and exits 0 (`AWS GetPublicKey error: service error`)


gal-fordefi:
### Summary
When signing an EIF at build time with a KMS key whose ARN belongs to a **different AWS account** than the calling IAM identity, `nitro-cli build-enclave` fails to read the signing key, prints a vague `Could not read signing info: "AWS GetPublicKey error: service error"`, and then **continues building an unsigned EIF while exiting 0**. `nitro-cli describe-eif` subsequently reports `"IsSigned": false`. The same KMS key/region/credentials work fine with `aws kms get-public-key`, and same-account signing works fine.

There are two distinct problems here:
1. Cross-account KMS keys can't be used for signing (the account is dropped from the key ARN before the `GetPublicKey` call), and the real SDK error is hidden behind a generic `service error`.
2. A signing failure silently degrades a *signed* build into an *unsigned* one with a success exit code.

### Environment
- `nitro-cli --version`: `Nitro CLI 1.4.4`
- OS: Amazon Linux 2023 (kernel `6.12.83-113.160.amzn2023.x86_64`)
- Region: `us-east-1`
- Build account (calling identity): `<BUILD_ACCOUNT>`
- KMS key account (different): `<KEY_ACCOUNT>`
- KMS key: `<KEY_ARN>` — asymmetric ECC signing key, with a cross-account key policy/grant allowing `kms:GetPublicKey` + `kms:Sign` to the build role.

### Steps to reproduce
1. Assume/run as an IAM role in `<BUILD_ACCOUNT>` that is granted `kms:GetPublicKey` and `kms:Sign` on a KMS key in a **different** account `<KEY_ACCOUNT>`.
2. Verify credentials and cross-account KMS access from the exact same shell/environment:
   ```bash
   aws sts get-caller-identity                                   # shows the expected assumed role in <BUILD_ACCOUNT>
   aws kms get-public-key --key-id <KEY_ARN> --region us-east-1   # SUCCEEDS, returns KeyId
   ```
3. Build a signed EIF with the same ARN:
   ```bash
   nitro-cli build-enclave \
     --docker-uri <IMAGE> \
     --private-key <KEY_ARN> \
     --signing-certificate <cert.pem> \
     --output-file app.eif
   ```
4. Inspect the result:
   ```bash
   nitro-cli describe-eif --eif-path app.eif   # "IsSigned": false
[...]

PR #757, opened 2026-08-05
Fail hard on sign_info setup issues

winkmw:
*Description of changes:*

This commit introduces a behavioural change to the error handling, to make the EIF building fail hard whenever it fails to prepare the signing info, and have it exit with a non-zero exit code.

Also fail hard if only providing one of the private key and certificate, since both are required to sign the EIF.

*Issue #, if available:*
#756 

### DIFF
--- a/enclave_build/src/lib.rs
+++ b/enclave_build/src/lib.rs
@@ -99,14 +99,11 @@ impl<'a> Docker2Eif<'a> {
         }
 
         let sign_info = match (private_key, certificate_path) {
-            (Some(key), Some(cert)) => SignKeyData::new(key, Path::new(&cert)).map_or_else(
-                |e| {
-                    eprintln!("Could not read signing info: {e:?}");
-                    None
-                },
-                Some,
+            (Some(key), Some(cert)) => Some(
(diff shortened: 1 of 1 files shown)


iorlov-dev (approved):

enclave_build/src/lib.rs
```diff
@@ -99,14 +99,11 @@ impl<'a> Docker2Eif<'a> {
...
-                |e| {
-                    eprintln!("Could not read signing info: {e:?}");
-                    None
-                },
-                Some,
+            (Some(key), Some(cert)) => Some(
+                SignKeyData::new(key, Path::new(&cert)).map_err(Docker2EifError::SignImageError)?,
             ),
-            _ => None,
+            (None, None) => None,
```
amateurcow, line 105: What about the case where neither are provided? Is the intention that it won't be signed in that case?

winkmw: Yes, the behavior is the same in that case (the `(None, None)` pattern), it just won't be signed, which is a valid thing to do.

amateurcow (approved):

